PR #34, opened 2017-09-06
Add new format based on the Coveralls one but with function information

### LINKED ISSUE #33: Add coveralls format output with function block information
As discussed on IRC, it would be nice to have an output format specifier that outputs coveralls format, but also adds function information to each child (esp. from where to where, line-wise, functions are in the source file).

This information is very useful for automated analysis on a per-function basis.

marco-c:
Fixes #33.

We can omit the digest here (and the token and commit sha as well?).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1502,7 +1502,7 @@ fn get_digest(path: PathBuf) -> String {
     }
 }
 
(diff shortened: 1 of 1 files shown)


choller:
Omitting the token and the source digest is great, we don't need it. However, we use the commit sha to read the revision out of the data. It is handy to have it in the data (we use it for the HG revision as well).

PR #65, opened 2018-01-08
Update README.md for the build & test

sylvestre:
I know this is trivial for rust developer but probably not for others...


marco-c:
The build failures are due to apt.llvm.org not being available...

PR #79, opened 2018-03-15
Add support for outputting branch coverage in LCOV format

### LINKED ISSUE #39: Support outputting branch coverage information in LCOV format
Currently we support outputting it in the Coveralls format, we should support the LCOV format too.

Sush0512:
Fixes #39.
Added the code for outputting the branch coverage data in `lcov` format
- As in the link  [geninfo](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php) given, added the code after the FNH as finished.

The case when there are no branches,
`BRF, BRH` values will be `0`

Thank you for the valuable guidance.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -122,7 +122,7 @@ pub fn output_lcov(results: CovResultIter) {
         // branch coverage information
         let mut branch_hit = 0;
         for (&(line, number), &taken) in &result.branches {
-            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken {"1"} else {"-"}).unwrap();
+            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken { "1" } else { "-" }).unwrap();
             if taken {
                 branch_hit = branch_hit + 1;
             }
--- a/src/output.rs
+++ b/src/output.rs
@@ -122,7 +122,7 @@ pub fn output_lcov(results: CovResultIter) {
         // branch coverage information
         let mut branch_hit = 0;
         for (&(line, number), &taken) in &result.branches {
-            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken {taken} else {"-"}).unwrap();
+            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken {"1"} else {"-"}).unwrap();
             if taken {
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -118,7 +118,19 @@ pub fn output_lcov(results: CovResultIter) {
             write!(writer, "FNF:{}\n", result.functions.len()).unwrap();
             write!(writer, "FNH:{}\n", result.functions.values().filter(|x| x.executed).count()).unwrap();
         }
-
+        
+        // branch coverage information
+        let mut branch_hit = 0;
+        for (&(line, number), &taken) in &result.branches {
+            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken {taken} else {"-"}).unwrap();
```
marco-c: This should be `if taken { "1" } else { "-" }`.

marco-c:
The build is failing, probably because of the issue I mentioned.

src/output.rs
```diff
@@ -118,7 +118,19 @@ pub fn output_lcov(results: CovResultIter) {
             write!(writer, "FNF:{}\n", result.functions.len()).unwrap();
             write!(writer, "FNH:{}\n", result.functions.values().filter(|x| x.executed).count()).unwrap();
         }
-
+        
+        // branch coverage information
+        let mut branch_hit = 0;
+        for (&(line, number), &taken) in &result.branches {
+            write!(writer, "BRDA:{},{},{},{}\n", line, 0, number, if taken {"1"} else {"-"}).unwrap();
```
marco-c: Nit: add spaces around "1" and "-", like this: `if taken { "1" } else { "-" }` to make the code more readable.

Sush0512:
I had a doubt regarding why the value of taken is `1`. Even the lcov file (/usr/lib/locv)` has used the taken value to be `taken`, as it represents a number indicating how many times the block has been taken.

marco-c:
> I had a doubt regarding why the value of taken is 1. Even the lcov file (/usr/lib/locv) has used the taken value to be taken, as it represents a number indicating how many times the block has been taken.

The parser only checks if a branch is taken or not and only stores a boolean variable saying if it was taken.
So for now we don't have enough information to tell how many times the branch was taken.

marco-c:
Thanks!

PR #81, opened 2018-03-29
Add tests for check_gcov_version

### LINKED ISSUE #63: Add tests for check_gcov_version
The function to test is in src/gcov.rs.

svensevenslow:
Fixes #63.

### DIFF
--- a/src/gcov.rs
+++ b/src/gcov.rs
@@ -95,4 +95,17 @@ mod tests {
         assert!(is_recent_version("gcov (Ubuntu 4.9.0-12ubuntu2) 4.9.0 20170406"));
         assert!(is_recent_version("gcov (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406"));
     }
+
+    #[cfg(unix)]
+    #[test]
+    fn test_check_gcov_version() {
+        check_gcov_version();    
+    }
+
+    #[cfg(windows)]
+    #[test]
+    #[should_panic]
+    fn test_check_gcov_version() {
+        check_gcov_version();
+    }
 }


src/gcov.rs
```diff
@@ -95,4 +95,9 @@ mod tests {
         assert!(is_recent_version("gcov (Ubuntu 4.9.0-12ubuntu2) 4.9.0 20170406"));
         assert!(is_recent_version("gcov (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406"));
     }
```
marco-c: Nit: newline between the two functions

src/gcov.rs
```diff
@@ -95,4 +95,9 @@ mod tests {
...
+    #[test]
+    fn test_check_gcov_version() {
+        assert!(check_gcov_version());
+    }
+
```
marco-c: Nit: no newline here

marco-c:
The test is failing on Windows as `gcov` is not installed there. We need to make the test you added not run on Windows and we need to add another test that only runs on Windows that checks that check_gcov_version fails.

We should mock `Command` to test these cases:
1) `gcov` doesn't exist;
2) `gcov` exists but fails to execute;
3) `gcov` returns an incompatible version;
4) `gcov` returns a compatible version.

The mocking can be done in a follow-up.

svensevenslow:
@marco-c  Hi! I don't understand how to make the test not run on Windows. The logic I used 
[here](https://github.com/marco-c/grcov/pull/81/files#diff-f8f2856508d959cac14df1d5770f78a9R98) is for the test to fail if gcov is not installed.

marco-c:
Here's an example of test that doesn't run on Windows: https://github.com/marco-c/grcov/blob/master/src/path_rewriting.rs#L148.
And here's an example of a test that only runs on Windows: https://github.com/marco-c/grcov/blob/master/src/path_rewriting.rs#L165.

svensevenslow:
@marco-c  I made these [changes](https://github.com/marco-c/grcov/pull/81/files#diff-f8f2856508d959cac14df1d5770f78a9R118), the test is still failing on Windows

marco-c:
> @marco-c I made these changes, the test is still failing on Windows

Check the logs and try to figure out why.

svensevenslow:
@marco-c Hi! I have updated the pr.

src/gcov.rs
```diff
@@ -95,4 +95,31 @@ mod tests {
         assert!(is_recent_version("gcov (Ubuntu 4.9.0-12ubuntu2) 4.9.0 20170406"));
         assert!(is_recent_version("gcov (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406"));
     }
+    
+    #[cfg(unix)]
+    #[test]
+    #[should_panic]
+    fn test_check_gcov_version_on_unix() {
+        let output = Command::new("gcov")
```
marco-c: This should just be like before, there should be no runtime check in the test.

svensevenslow:
@marco-c . Hope the pr is alright.

marco-c:
My earlier comment still stands:
> This should just be like before, there should be no runtime check in the test.

marco-c:
@svensevenslow are you still working on this?

svensevenslow:
@marco-c sorry for the delay, was busy with exams. I'll try to complete it

svensevenslow:
@marco-c When can gcov exist but fail to execute ? 
Also in order not to use an `assert` in the test, I was wondering if the correct way to go about it is this 

```#[cfg(unix)]
#[test]
fn test_check_gcov_version_on_unix() {
    //cases that will pass
    //check_gcov_version();
}

#[cfg(unix)]
#[test]
#[should_panic]
fn test_check_gcov_version_on_unix_should_fail() {
    //cases that wil fail
    //check_gcov_version();
}
```

marco-c:
> @marco-c When can gcov exist but fail to execute ?

If it crashes, but we can't make it crash on purpose. In order to test that you would need to mock "Command".We can do this in a follow-up though, for now we can just test that check_gcov_version returns true on Linux and panics on Windows.

svensevenslow:
@marco-c  I don't understand why this https://travis-ci.org/marco-c/grcov/jobs/383578218 is failing?

marco-c:
Don't worry about that, it's an unrelated failure.

PR #103, opened 2018-05-15
Merge files with same absolute path and different relative paths in the same bucket

### LINKED ISSUE #100: When parsing files, cleanup "." and ".." path components
For some files (luckily rarely and only for header files), in the reports we have things like "third_party/aom/aom_util/./aom_thread.h".
We can't just cleanup the paths at the end, as they appear both in the "third_party/aom/aom_util/aom_thread.h" and "third_party/aom/aom_util/./aom_thread.h" version in the same report.
We can clean them up when merging the results: https://github.com/marco-c/grcov/blob/aae34d4a234aeac63643676617f54b99c63e5601/src/lib.rs#L76.

calixteman:
Fixes #100.


src/main.rs
```diff
@@ -224,6 +225,12 @@ fn main() {
         prefix_dir = source_dir;
     }
 
+    let source_root = if source_dir != "" {
```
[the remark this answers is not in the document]
calixteman: we use the empty PathBuf as a PathBuf in path_rewriting and we just use it as an Option in lib::add_results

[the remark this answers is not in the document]
marco-c: Yeah, we should switch `rewrite_paths` to use an option too. But in the `rewrite_paths` case the code is clean as it is now as the `source_dir` is always used, instead in `add_results` you're doing `.to_str().unwrap().is_empty()` to check if the `source_dir` is empty, so using an option would be cleaner (and probably faster, since `add_results` is called hundred of thousands of times).

src/lib.rs
```diff
@@ -73,10 +73,21 @@ fn merge_results(result: &mut CovResult, result2: &mut CovResult) {
...
 
-fn add_results(mut results: Vec<(String,CovResult)>, result_map: &SyncCovResultMap) {
+fn add_results(mut results: Vec<(String,CovResult)>, result_map: &SyncCovResultMap, source_dir: &Option<PathBuf>) {
     let mut map = result_map.lock().unwrap();
     for mut result in results.drain(..) {
-        match map.entry(result.0) {
+        let path = if source_dir.is_some() {
+            // the goal here is to be able to merge results for paths like foo/./bar and foo/bar
+            let source_dir = source_dir.clone().unwrap();
+            match fs::canonicalize(source_dir.join(PathBuf::from(&result.0))) {
```
marco-c: Maybe you can do `source_dir.join(result.0)` directly, instead of turning result.0 into a PathBuf first?

src/lib.rs
```diff
@@ -73,10 +73,21 @@ fn merge_results(result: &mut CovResult, result2: &mut CovResult) {
...
 }
 
-fn add_results(mut results: Vec<(String,CovResult)>, result_map: &SyncCovResultMap) {
+fn add_results(mut results: Vec<(String,CovResult)>, result_map: &SyncCovResultMap, source_dir: &Option<PathBuf>) {
     let mut map = result_map.lock().unwrap();
     for mut result in results.drain(..) {
-        match map.entry(result.0) {
+        let path = if source_dir.is_some() {
+            // the goal here is to be able to merge results for paths like foo/./bar and foo/bar
+            let source_dir = source_dir.clone().unwrap();
```
marco-c: If you use match instead of if is_some maybe you can avoid cloning.

```Rust
let path = match source_dir {
    Some(source_dir) => {
      // ...
    },
    None => result.0
}
```

src/lib.rs
```diff
@@ -216,4 +227,35 @@ mod tests {
...
+        let cpp_file = fs::canonicalize(PathBuf::from("./test/relative_path/foo/bar/oof.cpp")).unwrap();
+        let cpp_file = cpp_file.to_str().unwrap();
+        let cov_result = result_map.get(cpp_file).unwrap();
+
+        assert_eq!(cov_result.lines, [(1,63), (2,63), (3,84), (4,42)].iter().cloned().collect());
+        assert!(cov_result.functions.contains_key("myfun"));
+    }
+
+    #[test]
+    fn test_ignore_relative_path() {
```
marco-c: Good idea adding a test for the None case too!

PR #110, opened 2018-05-19
Improve perf in passing gcno/gcda buffers into LLVM parser

### LINKED ISSUE #99: Pass gcno/gcda contents to the LLVM C++ parser directly, instead of letting it read files
This way, when we parse ZIP archives, we don't need to create gcno/gcda files, but we can pass their contents directly to the LLVM C++ parser.

calixteman:
Fixes #99.


src/c/llvmgcov.cpp
```diff
@@ -99,7 +98,27 @@ void parse_llvm_gcno(char* working_dir, char* file_stem, uint8_t branch_enabled)
...
   CustomGCOVFile GF;
+  std::string GCNO = std::string(file_stem) + ".gcno";
+
+  GCOVBuffer GCNO_GB(GCNO_Buff);
+  if (!GF.readGCNO(GCNO_GB)) {
+    errs() << "Invalid .gcno File!\n";
+    return;
+  }
+
+  GCOVBuffer GCDA_GB(GCDA_Buff);
```
marco-c: There are cases where there's only a gcno file and no gcda file (e.g. if a file is never executed). We should make sure this case is still supported.

src/producer.rs
```diff
@@ -119,28 +120,49 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
...
+                    gcno_buf = Some(Arc::new(buffer));
+                } else {
+                    extract_file(&mut gcno_file, &physical_gcno_path);
+                }
 
                 let gcda_path_in_zip = gcno_path_in_zip.with_extension("gcda");
 
                 for (num, &mut (gcda_archive_name, ref mut gcda_archive)) in gcda_archives.iter_mut().enumerate() {
                     let gcno_path = path.with_file_name(format!("{}_{}.gcno", stem, num + 1));
-
```
marco-c: Nit: leave this newline

src/producer.rs
```diff
@@ -166,15 +188,15 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
             queue.push(Some(WorkItem {
                 format: ItemFormat::INFO,
                 item: ItemType::Content(buffer),
-                name: archive_name.to_string(),
```
marco-c: Nit: leave the comma at the end

src/producer.rs
```diff
@@ -235,7 +257,10 @@ mod tests {
                     },
                     ItemType::Path(ref p) => {
                         elem.1 && p.ends_with(elem.2)
-                    }
+                    },
+                    ItemType::Buffers(_) => {
+                        true
```
marco-c: This should never happen, right? So return `false` here so that if it happens the assertion fails

src/producer.rs
```diff
@@ -119,28 +120,49 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
...
+                        if is_llvm {
+                            let mut gcda_buf: Vec<u8> = Vec::new();
+                            gcda_file.read_to_end(&mut gcda_buf).expect("Failed to read gcda file");
+                            let gcno_stem = path.with_file_name(format!("{}_{}", stem, num + 1));
+                            let gcno_stem = gcno_stem.to_str().expect("Failed to create stem file string");
+
+                            queue.push(Some(WorkItem {
+                                format: ItemFormat::GCNO,
+                                item: ItemType::Buffers(GcnoBuffers {stem: gcno_stem.to_string(),
+                                                                     gcno_buf: Arc::clone(&gcno_buf.clone().unwrap()),
```
marco-c: Let's try to use match to avoid cloning, as cloning the Option probably clones the buffer too: https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned.

src/producer.rs
```diff
@@ -119,34 +120,74 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
...
+                                }));
+                            }
+                        };
                     } else if num == 0 && !ignore_orphan_gcno {
-                        queue.push(Some(WorkItem {
-                            format: ItemFormat::GCNO,
-                            item: ItemType::Path(gcno_path),
-                            name: gcda_archive_name.to_string(),
-                        }));
+                        match gcno_buf_opt {
```
marco-c: It's a pity having to repeat this code twice, maybe we can refactor a bit to avoid it, but we can do it in the future.

calixteman: +1 let's do that later

PR #114, opened 2018-05-25
Improve perf in avoiding temp gcov file

### LINKED ISSUE #54: Make LLVM C++ parser directly populate the Rust result struct instead of generating gcov files
Right now, the Rust code is calling the C++ code to generate gcov files (in the intermediate format since ). Then, the Rust code parses the gcov files.

We should make this more intelligent, making the Rust code call the C++ code to parse the gcno file and the C++ code call back into Rust to populate the result struct while parsing.

Example of how to use a Rust object from C++: http://jakegoulding.com/rust-ffi-omnibus/objects/.

This 1) avoids I/O (as we wouldn't need to generate the gcov files), 2) avoids parsing the same info twice (once from the gcno and once from the gcov).

calixteman:
Fixes #54.


src/c/llvmgcov.cpp
```diff
@@ -10,30 +11,37 @@ class CustomFileInfo : public ProtectedFileInfo {
...
 };
 
-/// printIntermediate -  Print source files with collected line count information in the intermediate gcov format.
 void CustomFileInfo::printIntermediate(StringRef WorkingDir, StringRef MainFilename) {
-  std::string CoveragePath = getCoveragePath(MainFilename, MainFilename);
-  SmallString<128> FullCoveragePath(WorkingDir);
-  sys::path::append(FullCoveragePath, CoveragePath);
-  std::unique_ptr<raw_ostream> CovStream = openCoveragePath(FullCoveragePath);
-  raw_ostream &CovOS = *CovStream;
+    std::string CoveragePath = getCoveragePath(MainFilename, MainFilename);
```
marco-c: Nit: don't change the indentation here :P

src/lib.rs
```diff
@@ -121,25 +121,26 @@ pub fn consumer(working_dir: &PathBuf, source_dir: &Option<PathBuf>, result_map:
...
             ItemFormat::GCNO => {
                 let gcov_path = match work_item.item {
                     ItemType::Path(gcno_path) => {
                         if !is_llvm {
                             run_gcov(&gcno_path, branch_enabled, working_dir);
                         } else {
-                            call_parse_llvm_gcno(working_dir.to_str().unwrap(),
-                                                 gcno_path.parent().unwrap().join(gcno_path.file_stem().unwrap()).to_str().unwrap(),
-                                                 branch_enabled);
+                            result = Some(call_parse_llvm_gcno(working_dir.to_str().unwrap(),
```
marco-c: You can return here directly, so you don't need to have the Option result and you can clean up the rest of the code

src/lib.rs
```diff
@@ -121,25 +121,26 @@ pub fn consumer(working_dir: &PathBuf, source_dir: &Option<PathBuf>, result_map:
...
                         }
                         gcno_path.file_name().unwrap().to_str().unwrap().to_string() + ".gcov"
                     },
                     ItemType::Buffers(buffers) => {
-                        call_parse_llvm_gcno_buf(working_dir.to_str().unwrap(),
-                                                 &buffers.stem,
-                                                 &buffers.gcno_buf,
-                                                 &buffers.gcda_buf,
-                                                 branch_enabled);
+                        result = Some(call_parse_llvm_gcno_buf(working_dir.to_str().unwrap(),
```
marco-c: Same here

src/lib.rs
marco-c: Or we could split the is_llvm part and the non llvm part, as they are getting very different

PR #120, opened 2018-05-25
Check that llvm-config & clang++ versions are matching

calixteman:


tests/test.rs
```diff
@@ -241,18 +241,30 @@ fn check_equal_coveralls(expected_output: &str, output: &str, skip_branches: boo
...
+}
+
+fn get_compiler_major(version: &String) -> Result<String, &'static str> {
+    let re = Regex::new(r"(?:version |(?:gcc \([^\)]+\) )*)([0-9]+)\.[0-9]+\.[0-9]+").unwrap();
+    match re.captures(version) {
+        Some(caps) => {
+            Ok(caps.get(1).unwrap().as_str().to_string())
+        },
+        None => {
+            Err("Compiler version not found")
```
marco-c: You can just panic here, and make `get_compiler_major` return a `String`.

tests/test.rs
```diff
@@ -285,7 +297,9 @@ fn test_integration() {
             }
 
             println!("\nLLVM");
-            let llvm_version = get_version("llvm-config", "--version");
+            let llvm_version = get_version("llvm-config");
+            let clang_version = get_version("clang++");
+            assert_eq!(llvm_version, clang_version, "llvm-config ({:?}) and clang++ ({:?}) have not the same major version", llvm_version, clang_version);
```
marco-c: Nit: if you make `get_compiler_major` return a String, you can replace `{:?}` with `{}`.

Also `have not the same major version` -> `don't have the same major version`.

PR #121, opened 2018-05-27
Add unit tests with zip files for files generated with llvm

### LINKED ISSUE #116: Add integration test with a gcno zip and a gcda zip

### LINKED ISSUE #117: Add integration test with a gcno zip and a gcda zip containing gcno files with no linked gcda files

calixteman:
Fixes #116 and fixes #117.


calixteman:
The test is failing with COMPILE_VER=7 because there are 2 extra functions:
- _ZN4CiaoD1Ev => Ciao::~Ciao()
- _ZN4CiaoC1Ev => Ciao::Ciao()

which are not defined in class/main.cpp.
So a way to fix it could be to add these 2 functions in the code.
@marco-c wdyt ?

tests/test.rs
```diff
@@ -265,6 +271,38 @@ fn get_compiler_major(version: &String) -> String {
...
+            .expect(&format!("Cannot create zip for {:?}", zip_path));
+        zip.write_all(content.as_slice())
+            .expect(&format!("Cannot write {:?}", zip_path));
+    }
+    zip.finish()
+        .expect(&format!("Unable to write zip structure for {:?}", zip_path));
+}
+
 #[test]
 fn test_integration() {
```
marco-c: Let's add another test that only checks the zip stuff for a single directory (`tests/basic`), since testing on all of them is not needed. This way we can keep `test_integration` simpler and probably have a simpler test for gcno/gcda zip.

marco-c: Doing this will also fix the failure on Mac (since with `tests/basic` we don't have the additional two functions).

tests/class/expected_no_gcda_llvm.ade
```diff
@@ -0,0 +1,7 @@
+{"file":{"name":"main.cpp"},"language":"c/c++","method":{"covered":[],"name":"_ZN4CiaoD2Ev","percentage_covered":0.0,"total_covered":0,"total_uncovered":1,"uncovered":[6]}}
```
marco-c: You can remove all the expected files except the basic one now.

tests/test.rs
```diff
@@ -299,9 +339,59 @@ fn test_integration() {
...
+    create_zip(&gcda_zip_path, path, "*.gcda");
+    create_zip(&gcda0_zip_path, path, "");
+
+    let gcno_zip_path = path.join(gcno_zip_path);
+    let gcda_zip_path = path.join(gcda_zip_path);
+    let gcda0_zip_path = path.join(gcda0_zip_path);
+    let gcda1_zip_path = path.join(gcda1_zip_path);
+
+    // no gcda
+    check_equal_ade(&read_expected(path, "_no_gcda", "llvm", &llvm_version, "ade"),
```
marco-c: It might be better to use the `coveralls` format instead of `ade` for the test, so we also check the hit counts (they are not present in the `ade` format yet).

tests/test.rs
```diff
@@ -283,13 +321,15 @@ fn test_integration() {
 
             do_clean(path);
 
-            if !cfg!(windows) && !cfg!(target_os="macos") {
+            if false && !cfg!(windows) && !cfg!(target_os="macos") {
```
marco-c: Why did you add this `false`?

tests/test.rs
```diff
@@ -265,6 +271,38 @@ fn get_compiler_major(version: &String) -> String {
     }
 }
 
+fn create_zip(zip_path: &Path, base_dir: &Path, files_pat: &str) {
```
marco-c: Use `files_glob` instead of `files_pat`

tests/test.rs
```diff
@@ -57,7 +60,7 @@ fn run(path: &Path) {
     assert!(status.success());
 }
 
-fn read_expected(path: &Path, compiler: &str, compiler_ver: &str, format: &str) -> String {
+fn read_expected(path: &Path, append: &str, compiler: &str, compiler_ver: &str, format: &str) -> String {
```
marco-c: This function is slowly becoming very ugly, luckily it's a test so we don't care too much.

We could split it in two though, one function that only does:
```
     let mut f = File::open(path.join(&name)).expect(format!("{} file not found", name).as_str());
     let mut s = String::new();
     f.read_to_string(&mut s).unwrap();
     s
```

and the other which builds the path and then calls the first.

In test_integration we can call the second, in your test you can call the first with a hardcoded path.

marco-c:
I think the tests are failing because they are run in parallel and they both try to build grcov (as they run it via `cargo run`).
Two ways to solve this:
1) Figure out why `cargo run` rebuilds grcov (it shouldn't) and make it not rebuild it when it's already built;
2) Don't run `cargo run` as part of the test, but run the built grcov in the './target' directory.

marco-c:
> Figure out why cargo run rebuilds grcov (it shouldn't) and make it not rebuild it when it's already built;

The reason is https://github.com/alexcrichton/cc-rs/issues/230, but we'd have to wait for it to be fixed.

PR #128, opened 2018-05-28
Add integration test for zip stuff and gcc

### LINKED ISSUE #123: Make the basic_zip integration test also test gcc

calixteman:
Fixes #123.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,27 +14,10 @@ matrix:
             - llvm-4.0-dev
             - clang-4.0
             - llvm-4.0
-            - gcc-4.9
-            - g++-4.9
+            - gcc-5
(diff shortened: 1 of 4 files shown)


marco-c:
Still failing with GCC 4.9 on Linux: https://travis-ci.org/marco-c/grcov/jobs/384824416#L1057.
It looks like the actual result doesn't contain any file.

marco-c:
@calixteman you can either disable the test you've added for GCC 4.9, or disable GCC 4.9 testing entirely in .travis.yml.

marco-c:
(If you do the latter, please make sure there's at least one travis configuration where we test with Rust stable)

PR #129, opened 2018-06-07
Can read gcnos from a zip and gcdas from directory

### LINKED ISSUE #71: Support input of gcna/gcda files from a mix of ZIP archives and directories
At the moment, grcov supports reading gcno/gcda files from ZIP archives and directories, but it doesn't support reading gcno files from a ZIP archive and gcda files from a directory.

calixteman:
Fixes #71.


src/defs.rs
```diff
@@ -53,6 +57,76 @@ pub struct WorkItem {
     pub name: String,
 }
 
+#[derive(Debug)]
```
marco-c: If this stuff is only used in producer.rs, you can define everything there (`defs.rs` is for the shared stuff).

src/defs.rs
```diff
@@ -53,6 +57,76 @@ pub struct WorkItem {
...
+                match zip.by_name(name) {
+                    Ok(mut f) => {
+                        f.read_to_end(buf).expect("Failed to read gcda file");
+                        true
+                    },
+                    Err(_) => false,
+                }
+            },
+            ArchiveType::Dir(ref dir) => {
+                match File::open(dir.join(PathBuf::from(name))) {
```
marco-c: It might be possible to join with a string directly, instead of converting it to PathBuf first (I'm not sure).

src/defs.rs
```diff
@@ -53,6 +57,76 @@ pub struct WorkItem {
...
+                        true
+                    },
+                    Err(_) => false,
+                }
+            },
+            ArchiveType::Dir(ref dir) => {
+                let src_path = dir.join(PathBuf::from(name));
+                let dest_parent = path.parent().unwrap();
+                if !dest_parent.exists() {
+                    match fs::create_dir_all(dest_parent) {
```
marco-c: It should be faster if we try to create the directory without checking if it exists first, and then ignore the error if it already exists.

src/producer.rs
```diff
@@ -70,26 +70,10 @@ fn extract_file(zip_file: &mut zip::read::ZipFile, path: &PathBuf) {
     io::copy(zip_file, &mut file).expect("Failed to copy file from ZIP");
```
marco-c: You could remove `extract_file` if you turn gcno_archive into a `Archive` too.

src/producer.rs
```diff
@@ -216,26 +179,57 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
 }
 
 pub fn producer(tmp_dir: &Path, paths: &[String], queue: &WorkQueue, ignore_orphan_gcno: bool, is_llvm: bool) -> Option<Vec<u8>> {
-    let mut zip_files = Vec::new();
     let mut directories = Vec::new();
 
+    let mut gcno_archive: Option<ZipArchive<File>> = None;
+    let mut gcda_archives: Vec<(&String, Archive)> = Vec::new();
+    let mut info_archives: Vec<(&String, ZipArchive<File>)> = Vec::new();
```
marco-c: If we make `gcno_archive` and `info_archives` `Archive` objects too, we might be able to do some cleanups.

src/producer.rs
```diff
@@ -216,26 +179,57 @@ fn zip_producer(tmp_dir: &Path, zip_files: &[&String], queue: &WorkQueue, ignore
...
     }
 
-    let ret1 = zip_producer(tmp_dir, &zip_files, queue, ignore_orphan_gcno, is_llvm);
-    let ret2 = dir_producer(&directories, queue, ignore_orphan_gcno);
-
-    if ret1.is_some() {
-        ret1
-    } else if ret2.is_some() {
-        ret2
+    if gcno_archive.is_some() && gcda_archives.is_empty() && !directories.is_empty() {
```
marco-c: Do you think we could have a single producer that loops in all archives (where an "archive" would be either a Zip or a Directory) and generates data from all archives, instead of splitting it into two (zip_producer and dir_producer)?

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
+            });
         }
     }
 
-    let ret1 = zip_producer(tmp_dir, &zip_files, queue, ignore_orphan_gcno, is_llvm);
-    let ret2 = dir_producer(&directories, queue, ignore_orphan_gcno);
+    let gcnos: RefCell<HashMap<String, &Archive>> = RefCell::new(HashMap::new());
+    let gcdas: RefCell<HashMap<String, Vec<&Archive>>> = RefCell::new(HashMap::new());
+    let infos: RefCell<HashMap<String, Vec<&Archive>>> = RefCell::new(HashMap::new());
+    let linkeds: RefCell<HashMap<String, &Archive>> = RefCell::new(HashMap::new());
```
marco-c: Let's use a better name for this

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
 
-    for &mut (archive_name, ref mut archive) in &mut info_archives {
-        for i in 0..archive.len() {
-            let mut file = archive.by_index(i).unwrap();
-
-            if file.name().ends_with('/') {
-                continue;
-            }
-
+pub fn info_producer(infos: RefCell<HashMap<String, Vec<&Archive>>>, queue: &WorkQueue) {
```
marco-c: Maybe in the signature of the function we can use `&HashMap` instead of `RefCell<HashMap`?

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
+    for (name, archs) in infos.borrow().iter() {
```
marco-c: Nit: `archives` instead of `archs`, `archs` is confusing :P

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
 
-                let physical_gcno_path = path.with_file_name(format!("{}_{}.gcno", stem, 1));
+fn archive_producer(tmp_dir: &Path,
+                    gcnos: RefCell<HashMap<String, &Archive>>,
+                    gcdas: RefCell<HashMap<String, Vec<&Archive>>>,
+                    queue: &WorkQueue,
+                    ignore_orphan_gcno: bool,
+                    is_llvm: bool) {
+
+    for (stem, archive) in gcnos.borrow().iter() {
```
marco-c: `gcnos` -> `gcno_archives`

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
+                        let mut file = File::create(&path).expect("Failed to create file");
+                        io::copy(&mut f, &mut file).expect("Failed to copy file from ZIP");
+                        true
+                    },
+                    Err(_) => {
+                        false
+                    },
+                }
+            },
+            ArchiveType::Dir(ref dir) => {
```
marco-c: Why creating symlinks in this case? Can't we parse the files where they are directly?

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
+                       linkeds: &RefCell<HashMap<String, &'a Archive>>) {
+        match *self.item.borrow() {
+            ArchiveType::Zip(ref zip) => {
+                let mut zip = zip.borrow_mut();
+                for i in 0..zip.len() {
+                    let filename = zip.by_index(i).unwrap();
+                    let filename = filename.name();
+                    let path = PathBuf::from(filename);
+                    match path.extension() {
+                        Some(ext) => match ext.to_str().unwrap() {
```
marco-c: It's a pity that we have to replicate this code twice, can we share it somehow between the two branches?

src/producer.rs
calixteman: currently, run_gcov expects to have gcno and gcda in the same directory.
So if gcnos are in a zip and gcdas in a directory (or the opposite) then it won't work if we don't create symlinks.

marco-c: Ah right, gcov is a bit of a pain...

marco-c: We can create a hardlink with https://doc.rust-lang.org/std/fs/fn.hard_link.html.

calixteman: why is a hard link better than a soft link ?
Since we never modify gcno/gcda I don't see any interest to have a hard link.

marco-c: It simplifies the code, as it wouldn't require two different implementations for Windows and Linux

src/producer.rs
```diff
@@ -1,242 +1,343 @@
...
-                                let gcno_stem = path.with_file_name(format!("{}_{}", stem, num + 1));
-                                let gcno_stem = gcno_stem.to_str().expect("Failed to create stem file string");
+                    match gcno_buf_opt {
+                        Some(ref gcno_buf) => {
+                            let mut gcda_buf: Vec<u8> = Vec::new();
+                            let gcno_stem = tmp_dir.join(format!("{}_{}", stem, num + 1));
+                            let gcno_stem = gcno_stem.to_str().expect("Failed to create stem file string");
 
+                            if gcda_arch.read_in_buffer(&gcda, &mut gcda_buf) || (num == 0 && !ignore_orphan_gcno) {
                                 queue.push(Some(WorkItem {
```
marco-c: Maybe we can move the queue.push calls outside the match, so that we can share most of it?

src/producer.rs
calixteman: I just tried to replace soft links by hard ones and tests are failing on my machine because my base dir and the tmp dir are not on the same device.

src/producer.rs
calixteman: the two queue.push are preceeded by a "if gcda_archive.read_in_buffer..." and "if gcda_archive.extract(&gcda, &gcda_path)". The two conditions can be false because we don't find any gcda in archive which is ok when orphan are allowed. I could return an Option from the match but that means that we need to check if it's Some or not because pushing a None will stop everything (see lib::consumer). So I'm not sure we'll win something to put the queue.push outside the match.

src/producer.rs
```diff
@@ -148,13 +148,14 @@ impl Archive {
                 }
             },
             ArchiveType::Dir(ref dir) => {
+                // don't use a hard link here because it can fail when src and dst are not on the same device
                 let src_path = dir.join(name);
 
                 #[cfg(unix)]
-                os::unix::fs::symlink(&src_path, path).expect("Failed to create a symlink");
+                os::unix::fs::symlink(&src_path, path).expect(format!("Failed to create a symlink {:?} -> {:?}", src_path, path).as_str());
```
marco-c: I've always wondered if there isn't a better way to do this without format .as_str(), but I've never been able to find a nice alternative.

marco-c:
Landed in https://github.com/mozilla/grcov/commit/f20cd10704ca61f6d4c3058e7a4bb098aee44042.

PR #130, opened 2018-06-07
Enable some integration tests on Windows

marco-c:


marco-c:
We'll have to wait for Appveyor to update LLVM and Clang to 6 (right now they have Clang 6 and LLVM 5).

marco-c:
After #167, we might be able to do this.

Issue #92, opened 2018-04-19
Support parsing Jacoco artifacts


marco-c:
An example artifact is at https://github.com/marco-c/grcov/blob/master/test/basic-jacoco.xml.

gabriel-v:
Is [xml-rs](https://github.com/netvl/xml-rs) okay to use? It's the most popular one.

Your hints were:
- make [dir_producer](https://github.com/marco-c/grcov/blob/1311c289426a547c4c529809341ab5d7af2cfd7e/src/producer.rs#L12) find .xml files
- implement parsing those jacoco xml files, start from [parse_gcov](https://github.com/marco-c/grcov/blob/a06e7294b9e1e1cbb1c70f0a58bfb3012f2c6e2d/src/parser.rs#L320)
- wait for https://github.com/marco-c/grcov/pull/129 as it changes relevant portions

marco-c:
`xml-rs` sounds good!

> make dir_producer find .xml files

Both dir_producer and zip_producer (after #129 there should be a single one). Look at the `info` case, as that's the most similar one (for gcno/gcda we have couples of files, for info we have all the data in single files).

> implement parsing those jacoco xml files, start from parse_gcov

Yep, you can either look at parse_lcov or parse_gcov, they are the most simple formats.

> wait for #129 as it changes relevant logic

You can actually already start, the changes to `producer` will require some rebasing, but it should be very simple.

PR #136, opened 2018-06-25
Parse JaCoCo reports

gabriel-v:
Fixes https://github.com/mozilla/grcov/issues/92.

Still need to:

- [ ] find and pick up the `.xml` reports from JaCoCo
- [x] refactor the parser after review

This is my first rust patch, please review thoroughly


src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+}
+
+pub fn parse_jacoco_xml<T: Read>(xml_reader: BufReader<T>) -> Result<Vec<(String,CovResult)>, ParserError> {
+
+    let mut parser = EventReader::new(xml_reader);
+    let mut results = Vec::new();
+
+    loop {
+        match parser.next() {
+            Ok(XmlEvent::StartDocument { .. }) => {},
```
marco-c: Nit: this could go in the `_` case, since we're not doing anything

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+            Ok(XmlEvent::EndDocument) => break,
+            Ok(XmlEvent::StartElement { name, attributes, .. }) => {
+                match name.local_name.as_str() {
+                    "package" => {
+                        // Find package name
+                        let package = get_xml_attribute(&attributes, "name").unwrap();
```
marco-c: Let's use `except` here, so we can print a human-readable message (or we could return a ParserError, but since this is really unexpected I'd just use `except`).

gabriel-v: I could make `get_xml_attribute` to return `Result<>` and use `?` everywhere to propagate all the errors.

More or less, remove all `unwrap()` calls?

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+            },
+            _ => {},
+        }
+    }
+
+    Ok(functions)
+}
+
+fn parse_java_package<T: Read>(parser: &mut EventReader<T>, package: &str) -> Result<Vec<(String,CovResult)>, ParserError> {
+    let mut results_hash: HashMap<String, CovResult> = HashMap::new();
```
marco-c: Nit: let's call it results_map

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+                                    branches,
+                                });
+                            },
+                        };
+                    },
+                    &_ => {},
+                }
+            },
+            Ok(XmlEvent::EndElement{name}) => {
+                match name.local_name.as_str() {
```
marco-c: Nit: You can do a simple `if` here

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+            Err(e) => {
+                return Err(ParserError::Parse(e.to_string()))
+            },
+            _ => {}
+        }
+    }
+
+    Ok(
+        results_hash.into_iter().filter(
+            |(_, result)| result.lines.len() > 0 || result.branches.len() > 0
```
marco-c: Are there classes with no line information at all?

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+        ).map(
+            |(class, value)| (format!("{}/{}.java", package, class).trim_left_matches('/').to_string(), value)
```
marco-c: What are the initial '/' characters?

gabriel-v: There's at most one leading `/`, if the package is the empty string. The format will turn those into "/hello.java" instead of "hello.java".

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+                match name.local_name.as_str() {
+                    "counter" => {
+                        if get_xml_attribute(&attributes, "type").unwrap() == "METHOD" {
+                            executed = get_xml_attribute(&attributes, "covered").unwrap().parse::<u32>().unwrap() > 0;
+                        }
+                    },
+                    &_ => {},
+                }
+            },
+            Ok(XmlEvent::EndElement{ name }) => {
```
marco-c: You might be able to simplify this as something like:
```
Ok(XmlEvent::EndElement{ name }) if name.local_name.as_str() == "method" {
```

src/parser.rs
gabriel-v: Didn't find any classes with no line information. Instead, what I found were a few classes that were not the top class or inner class in their file. Those kind of classes are also removed by this filter.

This means we will need to figure out the mapping between java files and classes contained in them.

src/parser.rs
```diff
@@ -407,6 +411,243 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+                            for branch in 0..cb {
+                                branches.insert((nr, branch as u32), true);
+                            }
+
+                            for branch in cb..cb+mb{
+                                branches.insert((nr, branch as u32), false);
+                            }
+                        } else {
+                            // Line is a statement
+                            lines.insert(nr, ci);
```
marco-c: Is `ci` the number of times the instruction was executed?

gabriel-v: No, it's the count of covered instructions on that given code line.

marco-c:
Looks good overall! I think some of the `match` instructions could be cleaned up (either by adding the if in the outer match, or by using an if instead).
The `find and pick up the .xml reports from JaCoCo` can be done in a separate PR.

src/parser.rs
marco-c: Yes, that would work!

marco-c: So, use `?` when it's an error that could happen, and `expect` or `unwrap` when it something that is almost impossible.

src/parser.rs
marco-c: Oh that's a pity. An example?

src/parser.rs
marco-c: OK!

marco-c: Please add a comment

src/parser.rs
marco-c: OK. In Result, the lines map is `line number` -> `hit count`. We should have the same here.
Can you check with an example what happens in the XML file if you have a for loop that is executed multiple times?

src/parser.rs
gabriel-v: These are the files that contain multiple top-level java classes:

- [mobile/android/base/java/org/mozilla/gecko/RemotePresentationService.java](https://hg.mozilla.org/mozilla-central/file/0e0813df0f65/mobile/android/base/java/org/mozilla/gecko/RemotePresentationService.java)
        - org/mozilla/gecko/VirtualPresentation
- [mobile/android/base/java/org/mozilla/gecko/mdns/MulticastDNSManager.java](https://hg.mozilla.org/mozilla-central/file/0e0813df0f65/mobile/android/base/java/org/mozilla/gecko/mdns/MulticastDNSManager.java)
        - org/mozilla/gecko/mdns/DiscoveryListener
        - org/mozilla/gecko/mdns/DummyMulticastDNSManager
        - org/mozilla/gecko/mdns/RegistrationListener
        - org/mozilla/gecko/mdns/NsdMulticastDNSManager
        - org/mozilla/gecko/mdns/ResolveListener
        - org/mozilla/gecko/mdns/MulticastDNSEventManager

src/parser.rs
gabriel-v: [The JaCoCo docs](https://www.jacoco.org/jacoco/trunk/doc/counters.html) state:

> Due to the fact that a single line typically compiles to multiple byte code instructions the source code highlighting shows three different status for each line containing source code:

>  -  No coverage: No instruction in the line has been executed (red background)
>  -  Partial coverage: Only a part of the instruction in the line have been executed (yellow background)
>  -  Full coverage: All instructions in the line have been executed (green background)

> Depending on source formatting a single line of a source code may refer to multiple methods or multiple classes. 

It's not the count, but covered instructions per line.

marco-c: For reference, https://github.com/jacoco/jacoco/issues/89.

src/parser.rs
```diff
@@ -407,6 +418,196 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+
+                    for branch in cb..cb+mb{
+                        branches.insert((nr, branch as u32), false);
+                    }
+                } else {
+                    // This line is a statement.
+                    // JaCoCo does not feature execution counts, so we set the
+                    // count to 0 or 1.
+                    let hit = if ci > 0 { 1 } else { 0 };
+                    lines.insert(nr, hit );
```
marco-c: Nit: trailing space after `hit`

src/parser.rs
```diff
@@ -407,6 +418,196 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String,CovResult)>, ParserErr
...
+}
+
+pub fn parse_jacoco_xml_report<T: Read>(xml_reader: BufReader<T>) -> Result<Vec<(String,CovResult)>, ParserError> {
+    let mut parser = EventReader::new(xml_reader);
+    let mut results = Vec::new();
+
+    loop {
+        match parser.next() {
+            Ok(XmlEvent::StartElement {ref name, ref attributes, .. }) if name.local_name.as_str() == "package" => {
+                let package = get_xml_attribute(attributes, "name").unwrap();
```
marco-c: We can use ? rather than unwrap here

gabriel-v: Yeah, missed it. Thanks!

PR #137, opened 2018-06-27
Auto detect when a gcno has been produced with llvm

### LINKED ISSUE #25: Automatically identify LLVM gcno/gcda files by version
@jcranmer

calixteman:
Fix #25.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -17,10 +17,21 @@ pub enum ArchiveType {
     Dir(PathBuf),
 }
 
+pub enum FilePath<'a> {
+    File(&'a mut Read),
+    Path(&'a Path),
+}
+
 #[derive(Debug)]
 pub struct Archive {
     pub name: String,
     pub item: RefCell<ArchiveType>,
+}
+
+#[derive(Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


src/producer.rs
```diff
@@ -17,12 +17,23 @@ pub enum ArchiveType {
     Dir(PathBuf),
 }
 
+pub enum FilePath<'a> {
```
marco-c: We could probably get rid of this, if we make is_gcno_llvm accept a generic reader (e.g. like `parse_lcov` in src/parser.rs).
It's a minor thing though, so I've merged this and I'll file a follow-up 'good-first-bug' issue for it.

calixteman: is_gcno_llvm already take a std::io::Read, here I need to make the diff between Zip and Dir finally, since we've a Read when we're in a zip and we've a file when we're in a dir.

marco-c: The file could be read at https://github.com/mozilla/grcov/pull/137/files#diff-55015d4ff29cae5f493ea6ded47589b7R136, and the buffer passed to handle_file, so that in both cases (FilePath::File and FilePath::Path) you'd always pass a buffer to handle_file

PR #140, opened 2018-06-27
Add .gitignore for test output artifacts

gabriel-v:
Too much red in my `git status`.


calixteman:
Not sure that's a good idea because if we add some gcno, zip, ... in test directory we could miss them when adding new tests.

gabriel-v:
That's `test`, not `tests`.

calixteman:
oh my bad you're right

PR #141, opened 2018-06-27
Automatic formatting and style checking

gabriel-v:
Just to see how it looks like.

Apparently, the default is no trailing commas (or no commas at all?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,10 +86,12 @@ install:
     - ln -s `which gcc-${COMPILER_VER}` symlinks/gcc
     - ln -s `which g++-${COMPILER_VER}` symlinks/g++
     - ln -s `which gcov-${COMPILER_VER}` symlinks/gcov
+    - rustup component add rustfmt-preview
 script:
     - export PATH=$PWD/symlinks:$PATH
     - gcc --version
     - clang --version
+    - if [ -z "$TRAVIS_TAG" ] && [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo fmt --all -- --check; fi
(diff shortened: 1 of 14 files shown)


marco-c:
Overall the style looks nice to me. @calixteman do you have any other preference?

.rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+unstable_features = true
+indent_style = "Block"
+use_field_init_shorthand = true
+reorder_imports = false
```
marco-c: I would keep the reordering, sometimes it decreases the size of the patches.

calixteman: Looks good to me too: +1

.rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+unstable_features = true
+indent_style = "Block"
+use_field_init_shorthand = true
```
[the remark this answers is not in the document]
gabriel-v: https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#use_field_init_shorthand

Default: false
Stable: Yes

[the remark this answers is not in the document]
marco-c: OK, I'd prefer `false` here, so if that's the default let's leave it as `false`.

.rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+unstable_features = true
+indent_style = "Block"
```
marco-c: This is the default, right? So we can avoid setting it here?

gabriel-v: That's correct. The config file is now empty, so I'm removing it.

PR #149, opened 2018-07-02
Add --add-prefix argument to prepend a prefix to all file paths

### LINKED ISSUE #148: Option to prepend a path prefix
A command line argument to prepend a prefix to all file paths.

This is useful for getting the full relative paths of the source files referenced in the JaCoCo XML reports.

The JaCoCo XML reports only store the java package information, like `org/mozilla/gecko/browserid/DSACryptoImplementation.java`, but full paths are required: `mobile/android/app/src/test/java/org/mozilla/gecko/browserid/DSACryptoImplementation.java`.

Any preference on the option name? `--add-prefix`?

gabriel-v:
closes #148

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -1,7 +1,7 @@
 use globset::{Glob, GlobSetBuilder};
 use serde_json::Value;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use defs::*;
 use filter::*;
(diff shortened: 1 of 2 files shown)


src/path_rewriting.rs
```diff
@@ -56,8 +57,13 @@ pub fn rewrite_paths(
...
             path
         };
+        let path = path.replace("\\", "/");
 
-        let path = PathBuf::from(path.replace("\\", "/"));
+        let path = if let Some(prepend_path) = &prepend_prefix_dir {
+            Path::new(&prepend_path.to_str().unwrap().replace("\\", "/")).join(Path::new(&path))
+        } else {
+            PathBuf::from(path)
+        };
```
marco-c: You could do the replace("\\", "/") after, so you don't have to do it twice.

marco-c: And you can cleanup the first branch by directly joining prepend_path with path

PR #150, opened 2018-07-04
Remove unc paths coming from std::fs::canonicalize under windows

calixteman:
Under windows we've paths like "C:\\foo\\bar\\toto.c" and they aren't starting with "\\?C:\\foo\\bar" so that leads to consider that "C:\\foo\\bar\\toto.c" is not a file in the source_dir.


tests/test.rs
```diff
@@ -456,7 +456,7 @@ fn test_integration_zip() {
     for compiler in compilers {
         let is_llvm = compiler == "clang++";
         let name = if is_llvm { "llvm" } else { "gcc" };
-        let path = &PathBuf::from("tests/basic_zip");
+        let path = &PathBuf::from("tests/basic_zip_zip");
```
marco-c: we should update https://github.com/mozilla/grcov/pull/150/files#diff-dcd61c41ae35cab6b9646368cca02e49R391 to ignore this too.

tests/test.rs
```diff
@@ -538,3 +538,50 @@ fn test_integration_zip() {
...
+        println!("Integration tests still not supported under Windows.");
+        return;
+    }
+
+    let compilers = vec!["g++", "clang++"];
+
+    for compiler in compilers {
+        let is_llvm = compiler == "clang++";
+        let name = if is_llvm { "llvm" } else { "gcc" };
+        let path = &PathBuf::from("tests/basic_zip_dir");
```
marco-c: we should update https://github.com/mozilla/grcov/pull/150/files#diff-dcd61c41ae35cab6b9646368cca02e49R391 to ignore this too

src/path_rewriting.rs
```diff
@@ -16,6 +17,25 @@ fn to_uppercase_first(s: &str) -> String {
     c.next().unwrap().to_uppercase().collect::<String>() + c.as_str()
 }
 
+pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
+    let path = fs::canonicalize(path)?;
+
+    #[cfg(windows)]
+    let path = match {
```
gabriel-v: Why use `match` here? This also compiles:
```rust
    #[cfg(windows)]
    let path = {
        let spath = path.to_str().unwrap();
        if spath.starts_with(r"\\?\") {
            PathBuf::from(spath[r"\\?\".len()..].to_string())
        } else {
            path
        }
    };
```

calixteman: It compiles because you aren't under windows ;)

PR #154, opened 2018-07-06
Windows paths issue

calixteman:
Paths in zip are using "/" as separator and in a directory, paths under windows are using "\" so to fix the issue, all the "\" are replaced by "/" in paths and the matching between gcnos and gcdas is correct.


tests/test.rs
```diff
@@ -142,18 +156,8 @@ fn do_clean(directory: &Path) {
...
-    }
-    let to_remove_globset = glob_builder.build().unwrap();
-
-    for entry in WalkDir::new(&directory) {
-        let entry = entry.expect("Failed to open directory.");
-
-        if to_remove_globset.is_match(&entry.file_name()) {
-            fs::remove_file(entry.path()).unwrap();
-        }
+        rm_files(directory, to_remove_glob);
```
marco-c: You can pass to_remove_globs, so that you actually use the globset feature of having multiple globs in one

PR #157, opened 2018-07-12
Add early return for is_covered

### LINKED ISSUE #102: Add early return to `is_covered` when all lines are uncovered
Right now we check both `any_line_covered` and `any_function_covered` at the end.
We could return true early if `any_line_covered` is false.

#101 is needed first, so we can measure the improvement.

rhcu:
Fixes #102.

Benchmarks for is_covered:

**w/o early return:**
screen shot 2018-07-12 at 6 19 27 am

**with early return:**
screen shot 2018-07-12 at 6 23 56 am

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -7,6 +7,9 @@ pub fn is_covered(result: &CovResult) -> bool {
         .lines
         .values()
         .any(|&execution_count| execution_count != 0);
+    if any_line_covered == false {
+        return false;
+    }
     // For JavaScript files, we can't do the same, as the top-level is always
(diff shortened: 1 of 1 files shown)


src/filter.rs
```diff
@@ -7,6 +7,9 @@ pub fn is_covered(result: &CovResult) -> bool {
         .lines
         .values()
         .any(|&execution_count| execution_count != 0);
+    if any_line_covered == false {
```
marco-c: Since we are checking this earlier now, we shouldn't check it again at line 20.
Instead of `any_line_covered == false`, use `!any_line_covered`.

PR #164, opened 2018-09-21
[WIP] Compile grcov statically on Windows

marco-c:


marco-c:
I think this depends on https://github.com/mozilla/grcov/issues/166.

marco-c:
Actually this both depended and was necessary for #166, so I've moved the two commits there.

PR #167, opened 2018-09-26
Update to LLVM 7 and compile grcov statically on Windows

marco-c:
Fixes #166.

PR #169, opened 2018-09-27
Force installing Xcode-9.4 CLI tools

### LINKED ISSUE #165: Mac build is broken on Travis

marco-c:
Fixes #165.


marco-c:
Funny that Mac is now fixed, but Appveyor failed because the MSVC version was updated and so we can't find the redist DLLs anymore (luckily after #167 we won't need those DLLs anymore as we're going to build grcov statically).

PR #172, opened 2018-09-27
Filter jacoco and info files

calixteman:


calixteman:
Yes I think too.
Anyway afaik there are not so much jacoco files and if we don't do that it means that we need some extra steps to get gcno/gcda, correct jacoco files, ... so I think it's not a problem to lose few millisecs for that.

marco-c:
On CI we don't have any problem as we set GCOV_OUTPUT_DIR, so this is only going to slow down that use case.

calixteman:
So we could add an option "--check-files" or something like that. Wdyt ?

marco-c:
I'd say the perf hit is so small that we can ignore it.

PR #174, opened 2018-10-03
Add some do_clean in test.rs to avoid bad results when tests are failing

calixteman:


tests/test.rs
```diff
@@ -428,7 +428,7 @@ fn test_integration() {
 
             if !cfg!(windows) {
                 println!("GCC");
-                let gcc_version = get_version("gcc");
+                let gcc_version = get_version("g++");
```
marco-c: Why this change?

calixteman: I've several gcc versions on my computer and with this modif it's easier to replace g++ by g++-7 ;)

PR #175, opened 2018-10-03
Select compiler and gcov with env vars, disable GCC tests on Mac

calixteman:
When having several compilers, it's a pain to have to make symliks to select the correct one.
So with this patch we can easily select the good tools, e.g. "GRCOV_CXX=g++-7 GRCOV_GCOV=gcov-7 cargo test" will run tests with g++-7 instead of 8 which the default for g++ on my machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
         - GCOV=gcov-5
         - CLANG_CXX=clang++-7
         - LLVM_CONFIG=llvm-config-7
+        - DEPLOY=0
 
     - os: linux
       rust: stable
@@ -37,6 +38,7 @@ matrix:
         - GCOV=gcov-5
         - CLANG_CXX=clang++-8
         - LLVM_CONFIG=llvm-config-8
+        - DEPLOY=0
 
     - os: linux
       rust: nightly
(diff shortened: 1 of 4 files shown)


marco-c:
Sounds good to me, but we should unify the env variables from Travis.
On Travis we're currently using GCC_VER and CLANG_VER. Either use that ones for everything, or use the newly introduced ones for everything.

.travis.yml
```diff
@@ -126,6 +116,6 @@ deploy:
   on:
     rust: nightly
     tags: true
-    condition: $GCC_VER = 7
+    condition: $CLANG_CXX = clang++-7 OR $CLANG_CXX = clang++
```
marco-c: We should only deploy in one config for Linux and in one for Mac, with this condition we're deploying multiple times

marco-c:
You should now also be able to remove the expected files for GCC on Mac.

calixteman:
we don't have gcc files for mac only: find . -name "*gcc_mac*" return nothing

.travis.yml
calixteman: it was already the case before this patch, no ?

calixteman: but yes I agree

marco-c:
> we don't have gcc files for mac only: find . -name "gcc_mac" return nothing

Oh, nice :P

calixteman:
with the stars in the name

.travis.yml
calixteman: oh yes you're right

PR #179, opened 2018-10-04
Add code coverage to grcov

calixteman:


.travis.yml
```diff
@@ -112,6 +115,14 @@ script:
     - $CXX --version
     - if [ -z "$TRAVIS_TAG" ]; then cargo build --verbose; fi
     - if [ -z "$TRAVIS_TAG" ]; then cargo test -- --nocapture; fi
+    - |
+      if [ -z "$COVERAGE" ]; then
+        cargo clean;
+        CXXFLAGS="--coverage" LDFLAGS="--coverage -lgcov" CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1" cargo +nightly test;
```
marco-c: We can run the tests with these directly, so we don't need to run the tests twice

.travis.yml
```diff
@@ -112,6 +115,14 @@ script:
...
+        zip -0 ccov.zip `find . \( -name "grcov*.gc*" -o -name "llvmgcov.gc*" \) -print`;
+        ./target/debug/grcov ccov.zip -s src/ -t lcov --llvm --branch --ignore-dir libstd/* --ignore-dir libcore/* --ignore-dir liballoc/* --ignore-dir libtest/* --ignore-dir "<*" --ignore-dir "path_rewriting.rs" > lcov.info;
```
marco-c: Why do we need to ignore `path_rewriting.rs`?

.travis.yml
```diff
@@ -95,14 +95,17 @@ matrix:
         - GCOV=gcov-7
         - CLANG_CXX=clang++-7
         - LLVM_CONFIG=llvm-config-7
+        - COVERAGE=1
```
marco-c: Instead of defining COVERAGE, we can just upload coverage whenever we're using Rust nightly.

marco-c: This way we also merge the coverage we get in the different compiler configurations

.travis.yml
calixteman: It's strange here, in fact one gcno contains the path path_rewriting.rs and src/path_rewriting.rs.
So the latter is correct but not the former.

.travis.yml
```diff
@@ -103,15 +103,25 @@ matrix:
...
     - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew cask uninstall --force oclint; fi
     - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew install llvm@7 && brew link --overwrite --force llvm@7; fi
 script:
     - export CXX=$CLANG_CXX
     - $CXX --version
-    - if [ -z "$TRAVIS_TAG" ]; then cargo build --verbose; fi
-    - if [ -z "$TRAVIS_TAG" ]; then cargo test -- --nocapture; fi
+    - |
+      if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then
+        CXXFLAGS="--coverage" LDFLAGS="--coverage -lgcov" CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1" cargo +nightly build --verbose;
```
marco-c: Nit: you could do something like:
```
- if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then export ... ; fi
- cargo build --release
- cargo test --nocapture
- if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then PARSE AND UPLOAD; fi

.travis.yml
marco-c: Weird, maybe it's worth filing another bug in the Rust repository about this.

PR #181, opened 2018-10-04
Run tests before publishing releases too

### LINKED ISSUE #180: Run tests before publishing releases too
Right now, we're not running tests for builds triggered by a tag.

marco-c:
Fixes #180.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,25 +22,6 @@ matrix:
         - DEPLOY=0
 
     - os: linux
-      rust: stable
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,25 +21,6 @@ matrix:
         - LLVM_CONFIG=llvm-config-7
         - DEPLOY=0
 
-    - os: linux
```
calixteman: Is it expected to remove this build ?

marco-c: I've moved it at the end (and replaced rust stable with rust nightly like all the other builds except one), to put it closer to the other build which is similar to this one.
Not really needed, just a small change.

PR #194, opened 2018-10-08
Clean up path rewriting options

### LINKED ISSUE #76: Clean up path rewriting code
It was rewritten several times to cope with new corner cases. Now that everything works, I think the code should be rewritten from scratch to make it cleaner and to make it clearer what the path-mapping, prefix and source-dir options do by untangling them.

marco-c:
Fixes #76.

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -40,64 +40,59 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
 fn apply_mapping(mapping: &Option<Value>, path: &str) -> PathBuf {
     if let Some(mapping) = mapping {
         if let Some(p) = mapping.get(to_lowercase_first(path)) {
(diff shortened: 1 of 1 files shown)


src/path_rewriting.rs
```diff
@@ -45,72 +45,57 @@ pub fn rewrite_paths(
...
-            rel_path
-        } else if abs_path.starts_with(&source_dir) {
-            // Remove source dir from path.
-            abs_path.strip_prefix(&source_dir).unwrap().to_path_buf()
-        } else {
-            abs_path.clone()
-        };
+        if let Some(ref source_dir) = source_dir {
+            // Now the path could be different (if there was a symlink)
+            if abs_path.starts_with(&source_dir) {
```
calixteman: if source_dir is relative or if it's not canonicalized and abs_path is absolute then the condition should always be false

calixteman: but it won't never happen since we always pass a canonical path for source_dir (cf main.rs)

PR #197, opened 2018-10-08
Fix paths containg src dir

calixteman:


src/path_rewriting.rs
```diff
@@ -73,14 +90,14 @@ fn fixup_rel_path(source_dir: &Option<PathBuf>, abs_path: &PathBuf, rel_path: Pa
 }
 
 // Get the absolute path for the source file's path, resolving symlinks.
-fn get_abs_path(source_dir: &Option<PathBuf>, rel_path: PathBuf) -> (PathBuf, PathBuf) {
+fn get_abs_path(source_dir: &Option<PathBuf>, rel_path: PathBuf, cache: &mut Option<PathBuf>) -> (PathBuf, PathBuf) {
     let mut abs_path = if !rel_path.is_relative() {
         rel_path.clone()
     } else if let Some(ref source_dir) = source_dir {
         if !cfg!(windows) {
```
marco-c: Nit: you could do:
```
let rel_path = if !cfg!(windows) {
    rel_path
} else {
    PathBuf::from(&rel_path.to_str().unwrap().replace("/", "\\"))
}
guess_abs_path(&source_dir, &rel_path, cache)
```

calixteman: If I do that, then it doesn't compile since rel_path is moved and so no more usable in fixup call

marco-c: What if instead you do:
```Rust
let rel_path = if !cfg!(windows) {
    &rel_path
} else {
    &PathBuf::from(&rel_path.to_str().unwrap().replace("/", "\\"))
}
guess_abs_path(&source_dir, rel_path, cache)
```

(you could also change the name of `rel_path` if that makes it more readable)

calixteman: PathBuf::from.... doesn't live enough to get a ref on it

PR #202, opened 2018-10-10
Disable uploading the file network

marco-c:
Let's see if this removes the garbage we see at the beginning of the Appveyor coverage reports.


marco-c:
Still broken after this... Not sure what this feature is needed for, so I guess we can disable it anyways.

marco-c:
Actually, let's not disable it, let's keep the default options.

PR #217, opened 2018-10-25
Make source_dir cwd by default

mrbuzz:
This is my solution, it guarantees error handling on potential `env::current_dir()` failure and no impact on the rest of the code given that `source_dir` will be initialized with `&str`


calixteman:
The integration tests are failing, @mrbuzz could you investigate please ?

marco-c:
It looks like the integration tests are failing, you can run them locally by running `cargo test`.

src/main.rs
```diff
@@ -48,8 +48,9 @@ fn main() {
         print_usage(&args[0]);
         process::exit(1);
     }
+    let cwd = env::current_dir().expect("Failed to retrive current working directory");
+    let mut source_dir = cwd.as_os_str().to_str().unwrap();
```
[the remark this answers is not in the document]
mrbuzz: thanks for the review @calixteman maybe I can replace the `else ` branch starting at https://github.com/mozilla/grcov/blob/master/src/main.rs#L234 with the current working directory initialization. Sounds good?

mrbuzz:
Yes I'll investigate as soon as possible

marco-c:
Ah, getting the integration tests to work on Mac is not so easy :(

The fix for the test failure on CI is probably to just pass the source directory parameter always during tests, instead of only when the output format is `coveralls`: https://github.com/mozilla/grcov/blob/a49058409c493a49ce9a983aa9536a0341329069/tests/test.rs#L122.

mrbuzz:
@marco-c I made the change in `grcov/tests/test.rs` and now AppVeyor shows the same error that I obtain running tests locally as explained in the above comment https://github.com/mozilla/grcov/pull/217#issuecomment-433568040. I'd like to help more on that but I'd need more informations on how this works

marco-c:
To make the tests pass locally, set the `CONTINUOUS_INTEGRATION` environment variable (the tests are failing for you at https://github.com/mozilla/grcov/blob/a49058409c493a49ce9a983aa9536a0341329069/tests/test.rs#L283, simply because the line counts are not exactly the same because of a slightly different LLVM version, but this can be ignored and CONTINUOUS_INTEGRATION will do just that).

The problem you saw locally is about differing line counts, the problem on CI is another one:
```

**Comment by marco-c** [comment]
I think the fix I suggested is partial, you also need to make all calls to `run_grcov` in tests/test.rs pass a source_root (for example, https://github.com/mozilla/grcov/blob/a49058409c493a49ce9a983aa9536a0341329069/tests/test.rs#L456 is not passing it).

After that change, you might have to update the paths in the `expected` files.

**Comment by mrbuzz** [comment]
> I think the fix I suggested is partial, you also need to make all calls to `run_grcov` in tests/test.rs pass a source_root (for example,
> 
> [grcov/tests/test.rs](https://github.com/mozilla/grcov/blob/a49058409c493a49ce9a983aa9536a0341329069/tests/test.rs#L456)
> 
> Line 456 in [a490584](/mozilla/grcov/commit/a49058409c493a49ce9a983aa9536a0341329069)
>  &run_grcov(vec![path], &PathBuf::from(""), "ade"), 
> is not passing it).
> 
> After that change, you might have to update the paths in the `expected` files.

Thanks @marco-c! I updated the calls to `run_grcov` in `grcov/tests/test.rs` but I'm not sure on how to update the paths in the `expected` files. Could you provide me more information?

**Comment by mrbuzz** [comment]
> > I'm not sure on how to update the paths in the expected files. Could you provide me more information?
> 
> You can do that by simply looking at the failures of the integration tests. E.g. if you load the Travis build log, you see:
> 
> ```
> thread 'test_integration' panicked at 'Got unexpected {"file":{"name":"include.h"},"language":"c/c++","method":{"covered":[6],"name":"_ZN4CiaoC2Ev","percentage_covered":1.0,"total_covered":1,"total_uncovered":0,"uncovered":[]}} - Expected one of: {"file":{"name":"./include.h"},"language":"c/c++","method":{"covered":[6],"name":"_ZN4CiaoC2Ev","percentage_covered":1.0,"total_covered":1,"total_uncovered":0,"uncovered":[]}}
> {"file":{"name":"./include.h"},"language":"c/c++","method":{"covered":[6],"name":"_ZN4CiaoD2Ev","percentage_covered":1.0,"total_covered":1,"total_uncovered":0,"uncovered":[]}}
> {"file":{"name":"./include.h"},"language":"c/c++","method":{"covered":[11,12],"name":"_ZN4Ciao14calledFromFileEv","percentage_covered":1.0,"total_covered":2,"total_uncovered":0,"uncovered":[]}}
> {"file":{"covered":[6,11,12],"name":"./include.h","percentage_covered":1.0,"total_covered":3,"total_uncovered":0,"uncovered":[]},"is_file":true,"language":"c/c++","method":{"covered":[],"percentage_covered":null,"total_covered":0,"total_uncovered":0,"uncovered":[]}}
> {"file":{"name":"main.cpp"},"language":"c/c++","method":{"covered":[5,7,8,10,11],"name":"main","percentage_covered":1.0,"total_covered":5,"total_uncovered":0,"uncovered":[]}}
> {"file":{"covered":[5,7,8,10,11],"name":"main.cpp","percentage_covered":1.0,"total_covered":5,"total_uncovered":0,"uncovered":[]},"is_file":true,"language":"c/c++","method":{"covered":[],"percentage_covered":null,"total_covered":0,"total_uncovered":0,"uncovered":[]}}
> {"file":{"name":"file.cpp"},"language":"c/c++","method":{"covered":[4,5],"name":"_ZN4Ciao7setNameENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE","percentage_covered":1.0,"total_covered":2,"total_uncovered":0,"uncovered":[]}}
> {"file":{"name":"file.cpp"},"language":"c/c++","method":{"covered":[8,9],"name":"_ZN4Ciao7getNameB5cxx11Ev","percentage_covered":1.0,"total_covered":2,"total_uncovered":0,"uncovered":[]}}
> {"file":{"covered":[4,5,8,9],"name":"file.cpp","percentage_covered":1.0,"total_covered":4,"total_uncovered":0,"uncovered":[]},"is_file":true,"language":"c/c++","method":{"covered":[],"percentage_covered":null,"total_covered":0,"total_uncovered":0,"uncovered":[]}}
> ', tests/test.rs:227:9
> ```
> 
> ```
> {"file":{"name":"include.h"},"language":"c/c++","method":{"covered":[6],"name":"_ZN4CiaoC2Ev","percentage_covered":1.0,"total_covered":1,"total_uncovered":0,"uncovered":[]}}
> ```
> 
> is an unexpected entry, but you can actually see that there's an entry in the list which is exactly the same, but with a name being `./include.h` instead of `include.h`. So you'll just need to replace `./include.h` with `include.h` in the expected files that you find under the tests/ directory.
> 
> Are you able to run the integration tests now locally? This would speed up your work (otherwise you have to fix one path, then push to check if there's a new failure, then fix it, and so on).

Ah ok! Yes, now I'm able to run tests locally. I'll fix this asap

**Comment by mrbuzz** [comment]
@marco-c now only one job under Travis CI is failing. is it related to issue #213?

**Comment by marco-c** [comment]
> @marco-c now only one job under Travis CI is failing. is it related to issue #213?

Yes, I think so!

File: src/main.rs [hunk 2]
```diff
@@ -232,7 +232,8 @@ fn main() {
     let source_root = if source_dir != "" {
         Some(canonicalize_path(&source_dir).expect("Source directory does not exist."))
     } else {
-        None
+        let cwd = env::current_dir().expect("Failed to retrive current working directory");
```
marco-c: Typo here, this should read `retrieve`

marco-c:
We might have to add a new option to run grcov when you have no source directory (e.g. on mozilla-central test machines). @calixteman what do you think?

marco-c:
@mrbuzz would you mind adding the option for when users have no source directory? It should basically just set source_root to None.

mrbuzz:
> 
> 
> @mrbuzz would you mind adding the option for when users have no source directory? It should basically just set source_root to None.

@marco-c  Sure, I'll do it

mrbuzz:
Hey @marco-c! Apologize for the delay, let me know if there's something to modify.. should this new option be tested?

src/main.rs
```diff
@@ -230,11 +233,13 @@ fn main() {
     }
 
     let source_root = if source_dir != "" {
-        Some(canonicalize_path(&source_dir).expect("Source directory does not exist."))
-    } else {
-        let cwd = env::current_dir().expect("Failed to retrieve current working directory");
-        Some(canonicalize_path(&cwd.as_os_str().to_str().unwrap()).expect("Source directory does not exist."))
-    };
+            Some(canonicalize_path(&source_dir).expect("Source directory does not exist."))
```
marco-c: The indentation seems wrong here

marco-c:
>Hey @marco-c! Apologize for the delay, let me know if there's something to modify.. should this new option be tested?

No worries, and thanks again!

It would be good to test, but we can merge without a test too.

src/main.rs
mrbuzz: @marco-c this should be solved

marco-c:
It looks like an integration test is failing on Linux with LLVM 8. Maybe try just rebasing and seeing if the failure goes away?

Issue #207, opened 2018-10-10
Don't run travis CI when modifying appveyor.yml


calixteman:
https://github.com/facebook/react/pull/2000/files

marco-c:
And likewise, don't run Appveyor build when modifying .travis.yml.

sahitpj:
May I work on this issue, if not assigned?

marco-c:
Fixed by #218.

marco-c:
Reopening because #218 has been reverted.

rei2hu:
I looked at the revert reason. Is there a specific set of files where Travis should run but it doesn't? At a glance, it looks like this would only happen if each file changed contained `appveyor.yml` in its name.

marco-c:
> I looked at the revert reason. Is there a specific set of files where Travis should run but it doesn't? At a glance, it looks like this would only happen if each file changed contained appveyor.yml in its name.

I don't remember exactly when they were skipped, but IIRC they were skipped for any change. You can test by submitting a test PR applying the reverted patch and modifying some random files.

marco-c:
We are no longer using Appveyor.

PR #218, opened 2018-10-25
Don't build on Appveyor/Travis when changing Travis/Appveyor configuration

sahitpj:
Fixes issue #207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 sudo: false
 language: rust
 
-skip_commits:
-  files:
-    - appveyor.yml
+before_install:
+- |
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      TRAVIS_COMMIT_RANGE="FETCH_HEAD...$TRAVIS_BRANCH"
+    fi
+    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.appveyor.yml)' || {
+      echo "Only appveyor.yml was updated, stopping build process."
+      exit 
+    } 
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
     if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
       TRAVIS_COMMIT_RANGE="FETCH_HEAD...$TRAVIS_BRANCH"
     fi
-    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.appveyor.yml)' || {
+    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(appveyor.yml)' || {
       echo "Only appveyor.yml was updated, stopping build process."
       exit 
     } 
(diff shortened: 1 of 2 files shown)


marco-c:
Thanks for the patch! The same should be done for the other way around (when appveyor.yml is modified we shouldn't run Travis).
It looks like there's an error in the current patch though: `Error parsing appveyor.yml: Invalid "skip_commits" setting: file (Line: 6, Column: 3)`.

sahitpj:
Hey @marco-c will look into that immediately.

sahitpj:
screenshot 2018-10-25 16 13 57

hey @marco-c , it seems that only one build is failing, which I think that has something to do with another issue

calixteman:
@sahitpj you're right it's unrelated. So don't care about that.

.travis.yml
```diff
@@ -1,5 +1,10 @@
 sudo: false
 language: rust
+
+skip_commits:
```
marco-c: Are you sure that this is supported on Travis? I wasn't able to find any reference to this. If it isn't supported, then you'll have to do something like https://github.com/facebook/react/pull/2000/files.

sahitpj:
True, but I was wondering why `travis` didn't throw an error. Yeah I will implement that then @marco-c

sahitpj:
Hey @marco-c I think this should do

.travis.yml
```diff
@@ -1,5 +1,16 @@
 sudo: false
 language: rust
+
+before_install:
```
marco-c: Could you move this after `matrix` and before `install`? It's more coherent, as before_install happens on Travis right before install.

.travis.yml
```diff
@@ -1,5 +1,16 @@
...
+- |
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      TRAVIS_COMMIT_RANGE="FETCH_HEAD...$TRAVIS_BRANCH"
+    fi
+    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.appveyor.yml)' || {
```
marco-c: The appveyor file is `appveyor.yml`, not `.appveyor.yml`

.travis.yml
marco-c: Looks good, but this change is still missing @sahitpj

.travis.yml
```diff
@@ -1,5 +1,6 @@
 sudo: false
 language: rust
+
```
marco-c: Nit: no newline here

.travis.yml
```diff
@@ -103,7 +103,15 @@ matrix:
         - CLANG_CXX=clang++
         - LLVM_CONFIG=llvm-config
         - DEPLOY=1
-
```
marco-c: Nit: don't remove newline here :P

As a general rule, try to avoid unrelated whitespace changes (don't add/remove newlines in parts of the code you're not modifying), as they make the vcs history needlessly more complex.

This is OK though, don't worry about it (when a review comment starts with "Nit:" it means that it is a minor comment that can be also be ignored).

sahitpj: Will remember that, thanks a lot 😄

PR #219, opened 2018-10-25
Handle plain files as arguments: .info & .json only

calixteman:


src/producer.rs
```diff
@@ -218,6 +235,13 @@ impl Archive {
                     true
                 }
                 Err(_) => false,
+            }
+            ArchiveType::Plain(_) => match File::open(name) {
+                Ok(mut f) => {
+                    f.read_to_end(buf).expect("Failed to read gcda file");
```
marco-c: Nit: this is not only used for `gcda` files, right?

src/producer.rs
```diff
@@ -431,13 +471,31 @@ pub fn producer(
...
+                });
+            } else if let Some(ext) = full_path.clone().extension() {
+                let ext = ext.to_str().unwrap();
+                if ext == "info" || ext == "json" {
+                    plain_files.push(full_path);
+                } else {
+                    panic!("Cannot load file '{:?}': it isn't a .info or .json file.", full_path);
+                }
+            } else {
+                panic!("Cannot load file '{:?}': it isn't a .info or .json file.", full_path);
```
marco-c: Nit: `it isn't a directory, or a .info or a .json file`

src/producer.rs
```diff
@@ -196,6 +197,22 @@ impl Archive {
                     }
                 }
             }
+            ArchiveType::Plain(ref plain) => {
+                // All the paths are absolutes
+                for full_path in plain {
+                    //let path = PathBuf::from(full_path.file_name().unwrap());
```
marco-c: Nit: remove commented out statement

src/producer.rs
```diff
@@ -260,6 +284,20 @@ impl Archive {
 
                 true
             }
+            ArchiveType::Plain(_) => {
+                // don't use a hard link here because it can fail when src and dst are not on the same device
```
marco-c: We should never be in this branch, given that Plain is only for info files, right? If so, just remove all this code and add an assertion failing here

src/producer.rs
```diff
@@ -431,13 +471,31 @@ pub fn producer(
...
-                item: RefCell::new(ArchiveType::Dir(full_path)),
-            });
+            if full_path.is_dir() {
+                archives.push(Archive {
+                    name: path.to_string(),
+                    item: RefCell::new(ArchiveType::Dir(full_path)),
+                });
+            } else if let Some(ext) = full_path.clone().extension() {
+                let ext = ext.to_str().unwrap();
+                if ext == "info" || ext == "json" {
```
marco-c: We could handle xml files too

src/producer.rs
```diff
@@ -1412,6 +1470,29 @@ mod tests {
...
+    #[test]
+    fn test_plain_producer() {
+        let queue: Arc<WorkQueue> = Arc::new(MsQueue::new());
+
+        let tmp_dir = TempDir::new("grcov").expect("Failed to create temporary directory");
+        let tmp_path = tmp_dir.path().to_owned();
+        producer(
+            &tmp_path,
+            &[
+                "test/prova.info".to_string(),
```
marco-c: We should maybe test with a json and a xml too.

src/producer.rs
```diff
@@ -1412,6 +1470,29 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_plain_producer() {
```
marco-c: Also test the case where we have a gcno and a gcda and check that the function panics

src/producer.rs
calixteman: we could have several jacoco files coming from different machines

PR #220, opened 2018-10-25
naming threads as they are created

### LINKED ISSUE #216: Assign names to threads
We should assign names to threads when we create them, to make debugging easier when there is a problem.

LucianaMarques:
Fixes #216.

This pull request is to assign names to threads, as in [this issue](https://github.com/mozilla/grcov/issues/216).

I'm not sure, though, how to test this successfully!

I'm encountering errors while running on my computer.


marco-c:
> I'm encountering errors while running on my computer.

What errors exactly?

src/main.rs
```diff
@@ -283,7 +283,10 @@ fn main() {
         let working_dir = tmp_path.join(format!("{}", i));
         let source_root = source_root.clone();
 
-        let t = thread::spawn(move || {
+        let thread_name = String::from("");
+        thread_name.push(i.to_string());
+            
+        let t = thread::::Builder::new().name(thread_name).spawn(move || {
```
marco-c: There's also another thread spawned a few lines earlier to which we should assign a name.

src/main.rs
```diff
@@ -283,7 +283,10 @@ fn main() {
         let working_dir = tmp_path.join(format!("{}", i));
         let source_root = source_root.clone();
 
-        let t = thread::spawn(move || {
+        let thread_name = String::from("");
+        thread_name.push(i.to_string());
```
marco-c: The name should be meaningful, only assigning an integer will not improve the situation much. We should use a name like `producer-` + i.

LucianaMarques:
> What errors exactly?

When I type "cargo build" in the command line, the following message appears (doesn't look like to be due to the changes I did):

error: failed to run custom build command for `grcov v0.3.2 (file:///home/luciana/Documents/Mozilla/grcov)`
process didn't exit successfully: `/home/luciana/Documents/Mozilla/grcov/target/debug/build/grcov-303159f620a597f4/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Error while running llvm-config: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

warning: build failed, waiting for other jobs to finish...
error: build failed

src/main.rs
LucianaMarques: Ok! I was thinking about this too, I'll add it now. But I'm not sure this code compiles either with my modifications, once I get the program running I'll correct this.

marco-c:
In order to build, you will need to install the LLVM 7 libraries and headers. That error is showing up because they are missing (in particular, it can't find the `llvm-config` program).

If you are using Linux, try installing LLVM 7 using your package manager. If you're using Windows, you can use the LLVM taken from https://github.com/marco-c/llvm-appveyor/releases/tag/llvm_342383.

LucianaMarques:
>  LLVM 7

Hi Marco, I installed everything from the qualification branch but it's still not working. Do you have any idea of what else it could be? Thanks

LucianaMarques:
Hey, 

I just made another commit. I'm finding this error here, though, and I'm not sure if it's related to my work:

error[E0599]: no method named `join` found for type `std::result::Result<std::thread::JoinHandle<()>, std::io::Error>` in the current scope
   --> src/main.rs:311:16
    |
311 |         parser.join().unwrap();
    |                ^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `grcov`.

marco-c:
> I just made another commit. I'm finding this error here, though, and I'm not sure if it's related to my work:

Yeah I think it is, to confirm you can just try to compile without your changes. If the compilation works without your changes and doesn't with your changes, then the error is due to your changes.

LucianaMarques:
> > I just made another commit. I'm finding this error here, though, and I'm not sure if it's related to my work:
> 
> Yeah I think it is, to confirm you can just try to compile without your changes. If the compilation works without your changes and doesn't with your changes, then the error is due to your changes.

Ok, I used [this reference](https://doc.rust-lang.org/std/thread/struct.Builder.html) to try to fix it.

LucianaMarques:
The appveyor test failed, but the project is compiling in my computer.

marco-c:
> The appveyor test failed, but the project is compiling in my computer.

There's something strange in your patch, it contains `srcmain.rs:311:16`.

src/main.rs
```diff
@@ -283,7 +283,12 @@ fn main() {
         let working_dir = tmp_path.join(format!("{}", i));
         let source_root = source_root.clone();
 
-        let t = thread::spawn(move || {
+        let mut  thread_name = String::from("Consumer ");
+        thread_name.push_str(&i.to_string());
```
marco-c: You can use `format!`, which would make creating the name nicer (no need for a push)

src/main.rs
```diff
@@ -283,7 +283,12 @@ fn main() {
...
+
+        let t = thread::Builder::new().name(thread_name);
```
marco-c: Try to see if you can avoid defining the name in a separate variable and just use it here

src/main.rs
```diff
@@ -283,7 +283,12 @@ fn main() {
...
+        
+        let join_handle_t = t.spawn(move || {
```
marco-c: Try to see if you can do `thread::Builder::new().name(thread_name).spawn` instead

LucianaMarques:
> > The appveyor test failed, but the project is compiling in my computer.
> 
> There's something strange in your patch, it contains `srcmain.rs:311:16`.

Yeap, really strange, I may have accidentally created a file. It was empty and the project compiled without it, so I just removed it.

src/main.rs
LucianaMarques: I was doing this before and join() wasn't working. I understood that threads created with Builder cannot join().

marco-c: I think they can, it's just that you needed an `unwrap` at the end of `spawn` to actually get a JoinHandle object, otherwise you get a [Result](https://doc.rust-lang.org/std/result/) object.

src/main.rs
LucianaMarques: Done!

src/main.rs
LucianaMarques: Done!

src/main.rs
LucianaMarques: Done!

LucianaMarques:
@marco-c hey marco, only one test in travis failed. I tried to take a look at it but didn't understand much. I'm not sure if it's due to what I changed, we basically only added names to the threads, now it's almost as it was before I started working. If you can please tell me more about what you're checking on this tests! Thanks

src/main.rs
```diff
@@ -283,7 +283,7 @@ fn main() {
         let working_dir = tmp_path.join(format!("{}", i));
         let source_root = source_root.clone();
 
-        let t = thread::spawn(move || {
+        let t = thread::Builder::new().name(format!("Consumer {}",&i.to_string())).spawn(move || {
```
marco-c: Nit: space between the parameters of the function: `format!("Consumer {}", &i.to_string())`.

marco-c: You probably don't need to turn `i` into a string, format! should be able to handle integers directly

marco-c:
> @marco-c hey marco, only one test in travis failed. I tried to take a look at it but didn't understand much. I'm not sure if it's due to what I changed, we basically only added names to the threads, now it's almost as it was before I started working. If you can please tell me more about what you're checking on this tests! Thanks

Don't worry about that one, it is a known failure (#213).

The patch looks almost ready to land now, still need to add the name to the other thread.

LucianaMarques:
> > @marco-c hey marco, only one test in travis failed. I tried to take a look at it but didn't understand much. I'm not sure if it's due to what I changed, we basically only added names to the threads, now it's almost as it was before I started working. If you can please tell me more about what you're checking on this tests! Thanks
> 
> Don't worry about that one, it is a known failure (#213).
> 
> The patch looks almost ready to land now, still need to add the name to the other thread.

Great, thank you! I'll make the final modifications today.

src/main.rs
LucianaMarques: You mean I need to remove the space?

marco-c: I mean you need to add the space (there's currently no space between the two parameters of format!)

LucianaMarques: Done! Compiled without turning into string, indeed.

src/main.rs
```diff
@@ -254,7 +254,7 @@ fn main() {
         let path_mapping_file = path_mapping_file.to_owned();
         let path_mapping = Arc::clone(&path_mapping);
 
-        thread::spawn(move || {
+        thread::Builder::new().name(format!("Producer Thread")).spawn(move || {
```
marco-c: Let's remove ` Thread`, as it's implied already

src/main.rs
```diff
@@ -283,7 +283,7 @@ fn main() {
         let working_dir = tmp_path.join(format!("{}", i));
         let source_root = source_root.clone();
 
-        let t = thread::spawn(move || {
+        let t = thread::Builder::new().name(format!("Consumer {}",&i)).spawn(move || {
```
marco-c: Don't forget the space between the parameters of format! :smile: 
You might not need the & operator before i.

LucianaMarques:
> > @marco-c hey marco, only one test in travis failed. I tried to take a look at it but didn't understand much. I'm not sure if it's due to what I changed, we basically only added names to the threads, now it's almost as it was before I started working. If you can please tell me more about what you're checking on this tests! Thanks
> 
> Don't worry about that one, it is a known failure (#213).
> 
> The patch looks almost ready to land now, still need to add the name to the other thread.

I added a name, but would like to confirm whether it's a relevant one.

LucianaMarques:
My latest patch produces the following error. I'm creating the thread's name as previously, and can't find much about this problem in the internet. What could be wrong?

error[E0599]: no method named `join` found for type `()` in the current scope
   --> src/main.rs:301:22
    |
301 |     let _ = producer.join();
    |                      ^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `grcov`.

marco-c:
> error[E0599]: no method named join found for type () in the current scope

This means `producer` is of type `()`, while it should be of type Thread.

producer is defined at https://github.com/mozilla/grcov/pull/220/files/faa0daf822491075cab6e1842336ac046ef65485..f072ad0eba0fddc1d6e3398b6d20f8729c7def6b#diff-639fbc4ef05b315af92b4d836c31b023R251. In your change, the if expression is no longer returning a Thread but a `()` because you're ending the statement with `;`.

LucianaMarques:
> producer is defined at https://github.com/mozilla/grcov/pull/220/files/faa0daf822491075cab6e1842336ac046ef65485..f072ad0eba0fddc1d6e3398b6d20f8729c7def6b#diff-639fbc4ef05b315af92b4d836c31b023R251. In your change, the if expression is no longer returning a Thread but a `()` because you're ending the statement with `;`.

Thank you! Now it's working! :)

src/main.rs
```diff
@@ -254,7 +254,7 @@ fn main() {
         let path_mapping_file = path_mapping_file.to_owned();
         let path_mapping = Arc::clone(&path_mapping);
 
-        thread::spawn(move || {
+        thread::Builder::new().name(format!("Producer")).spawn(move || {
```
marco-c: No need for format! here, since this is not a formatting string (it doesn't contain any `{}`)

LucianaMarques: done

PR #222, opened 2018-10-28
Exit with 1 when a thread is panicking

### LINKED ISSUE #215: If a thread fails, the error code should not be 0
```
marco@marco ~/Documenti $ grcov jsvm/file.info 
thread '<unnamed>' panicked at 'No input files found', src/producer.rs:425:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
marco@marco ~/Documenti $ echo $?
0
```

calixteman:
Fixes #215.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -297,11 +297,8 @@ fn main() {
         parsers.push(t);
     }
 
-    match producer.join() {
-        Ok(_) => { }
-        Err(_) => {
-            process::exit(1);
-        }
+    if let Err(_) = producer.join() {
+        process::exit(1);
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -297,15 +297,25 @@ fn main() {
         parsers.push(t);
     }
 
-    let _ = producer.join();
+    match producer.join() {
```
marco-c: Nit: you could use if-let in cases like this:
```Rust
if let Err(_) = producer.join() {
  ...
}
```

PR #223, opened 2018-10-30
Sort ignore_dirs before to build GlobSet

calixteman:


src/path_rewriting.rs
```diff
@@ -175,10 +175,11 @@ pub fn rewrite_paths(
     source_dir: Option<PathBuf>,
     prefix_dir: Option<PathBuf>,
     ignore_not_existing: bool,
-    to_ignore_dirs: Vec<String>,
+    mut to_ignore_dirs: Vec<String>,
```
marco-c: It would be nice not to require mutation, but I guess the only way would be to clone the Vec (or use an iterator, but it would either require a complex implementation or using a third-party library such as https://docs.rs/itertools/0.5.8/itertools/trait.Itertools.html#method.sorted).
So, let's keep the mut.

calixteman: My first idea was to insert-sort the vec when building it in main.rs and so not to have this "mut" here.

PR #230, opened 2018-11-18
Reader

calixteman:
- write a rust GCNO/GCDA parser
- write a rust GCOV stuff (cycle counts and so on)
- for now it isn't plugged at all (let's do that in another PR)


src/reader.rs
```diff
@@ -0,0 +1,939 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    
+    use super::*;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_reader_gcno() {
```
marco-c: Many parts of these tests are shared, maybe you can refactor them out in a function that you reuse in all tests?

src/reader.rs
```diff
@@ -0,0 +1,939 @@
...
+                    reader.read_u32::<LittleEndian>()?;
+                    reader.read_u64::<LittleEndian>()?;
+                    self.runcounts += reader.read_u32::<LittleEndian>()?;
+                }
+                Ok(())
+            }
+        }
+    }
+
+    fn read_gcda_function(version: u32, checksum: u32, fun: &mut GcovFunction, reader: &mut Read) -> Result<(), GcovError> {
```
marco-c: Try to return the `Err`s early in this function, so you can reduce the amount of indentation.

marco-c:
It would be really interesting to write some tests using Hypothesis (https://hypothesis.readthedocs.io/en/latest/), matching the output of the new reader with the output of the LLVM-based reader, but it might be quite difficult.

src/lib.rs
```diff
@@ -174,19 +184,19 @@ pub fn consumer(
...
-                        drop(buffers.gcno_buf);
-                        result
+                        match GCNO::compute(&buffers.stem,
+                                            buffers.gcno_buf,
+                                            buffers.gcda_buf,
+                                            branch_enabled) {
+                            Ok(r) => r,
+                            Err(e) => {
+                                // Just print the error, don't panic and continue
+                                eprintln!("{}", e);
```
marco-c: Add an error message here before `{}`, so when it happens we can tell where it comes from

src/lib.rs
```diff
@@ -51,18 +54,25 @@ fn merge_results(result: &mut CovResult, result2: &mut CovResult) {
...
     }
 
-    for (&(line_no, number), &taken) in &result2.branches {
-        match result.branches.entry((line_no, number)) {
+    for (line_no, taken) in result2.branches {
+        match result.branches.entry(line_no) {
             btree_map::Entry::Occupied(c) => {
-                *c.into_mut() |= taken;
+                let v = c.into_mut();
+                for (x, y) in taken.iter().zip(v.iter_mut()) {
```
marco-c: You're handling the case where taken has more elements than v, but what if the opposite happens?

src/output.rs
```diff
@@ -130,9 +130,12 @@ pub fn output_lcov(results: CovResultIter) {
         // println!("{} {:?}", rel_path, result.lines);
 
         writeln!(writer, "SF:{}", rel_path.display()).unwrap();
-
+        let mut debug = false;
```
marco-c: I guess you wanted to remove this debug stuff?

src/reader.rs
```diff
@@ -0,0 +1,939 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    fn collect_lines(&self) -> HashMap<String, HashMap<u32, u64>> {
+        let mut results: HashMap<String, HashMap<u32, u64>> = HashMap::new();
+        for function in &self.functions {
+            // TODO: could we remove the clone ?
```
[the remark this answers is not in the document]
calixteman: collect_lines is used in dump which is itself used in test of for debugging purposes.
So if used to debug then I think we shouldn't remove the functions.

src/lib.rs
calixteman: if len(v) >= len(taken) then only the n (=len(taken)) elements of v are modified

src/reader.rs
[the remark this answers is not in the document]
marco-c: Maybe add another function for debugging purposes and keep the "production" one as fast as possible.

src/lib.rs
marco-c: :+1:

appveyor.yml
```diff
@@ -30,8 +30,6 @@ install:
   - 7z x clang.tar.bz2
   - 7z x clang.tar
   - set PATH=clang\bin;%PATH%
-  - llvm-config --version
```
marco-c: We don't need to download Clang anymore, right? You could remove the previous lines too if we don't.

.travis.yml
```diff
@@ -18,7 +18,6 @@ matrix:
         - GCC_CXX=g++-5
         - GCOV=gcov-5
         - CLANG_CXX=clang++-7
-        - LLVM_CONFIG=llvm-config-7
```
marco-c: Same for Travis, we might not need Clang at all anymore.

src/output.rs
```diff
@@ -130,7 +130,6 @@ pub fn output_lcov(results: CovResultIter) {
         // println!("{} {:?}", rel_path, result.lines);
 
         writeln!(writer, "SF:{}", rel_path.display()).unwrap();
-
```
marco-c: Nit: no need to remove this line

appveyor.yml
calixteman: we need to have clang for the integration test

marco-c: Ah, right!

PR #233, opened 2018-12-09
Be more verbose when gcov fails

### LINKED ISSUE #232: When gcov fails, output its stdout and stderr

marco-c:
Since I was here, I've also removed some useless commented-out code (in a separate commit).

Fixes #232.

### DIFF
--- a/src/gcov.rs
+++ b/src/gcov.rs
@@ -3,37 +3,6 @@ use std::env;
 use std::path::PathBuf;
 use std::process::Command;
 
-/*
-#[link(name = "gcov")]
-extern {
-    fn __gcov_read_unsigned() -> u32;
-    fn __gcov_open(name: *const c_char) -> i32;
-    fn __gcov_close();
-}
-
-fn gcov_open(file: String) -> i32 {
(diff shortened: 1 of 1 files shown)


src/gcov.rs
```diff
@@ -51,21 +20,16 @@ pub fn run_gcov(gcno_path: &PathBuf, branch_enabled: bool, working_dir: &PathBuf
...
-        gcno_path.display()
-    );
-    //}
+    let output = status.output().expect("Failed to execute gcov process");
+
+    if !output.status.success() {
+        eprintln!("gcov stdout: {}", String::from_utf8_lossy(&output.stdout));
+        eprintln!("gcov stderr: {}", String::from_utf8_lossy(&output.stderr));
+
+        panic!("gcov wasn't successfully executed on {}", gcno_path.display());
```
sylvestre: maybe improve the wording with 
gcov execution failed on {}

marco-c: Good idea, done with dba3edcf581a6f46e13d4ec4d2086809b87fc0d6.

PR #236, opened 2018-12-29
Move from tempdir to tempfile

sylvestre:


Cargo.toml
```diff
@@ -37,7 +37,7 @@ num_cpus = "^1.8"
 semver = "^0.9"
 rust-crypto = "^0.2"
 zip = { version = "^0.4", features = ["deflate-zlib"], default-features = false }
-tempdir = "^0.3"
+tempfile = "3"
```
marco-c: Use `^` so we are more relaxed with the requirement.

src/main.rs
```diff
@@ -241,8 +240,9 @@ fn main() {
         Some(PathBuf::from(prefix_dir))
     };
 
-    let tmp_dir = TempDir::new("grcov").expect("Failed to create temporary directory");
+    let tmp_dir = tempfile::tempdir().unwrap();
```
marco-c: Keep `expect` instead of `unwrap` so we have an error message should this fail (`except` is basically the same as `unwrap`, but with an error message).

src/producer.rs
```diff
@@ -614,7 +615,7 @@ mod tests {
     fn test_dir_producer() {
         let queue: Arc<WorkQueue> = Arc::new(MsQueue::new());
 
-        let tmp_dir = TempDir::new("grcov").expect("Failed to create temporary directory");
+        let tmp_dir = tempfile::tempdir().unwrap();
```
marco-c: Same here and in the other cases (`except` instead of `unwrap`).

PR #240, opened 2019-01-14
Update the zip crates version to use the latest

sylvestre:
Not sure it will work

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ walkdir = "^2.2"
 num_cpus = "^1.10"
 semver = "^0.9"
 md-5 = "^0.8"
-zip = { version = "^0.4", features = ["deflate-zlib"], default-features = false }
+zip = { version = "^0.5", features = ["deflate-zlib"], default-features = false }
 tempfile = "^3"
 uuid = { version = "^0.7", features = ["v4"] }
 globset = "^0.4"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -36,7 +36,7 @@ walkdir = "^2.2"
 num_cpus = "^1.8"
 semver = "^0.9"
 md5 = "^0.6"
-zip = { version = "^0.4", features = ["deflate-zlib"], default-features = false }
+zip = "^0.5"
```
marco-c: I used the features thing so it would reduce the size of the resulting binary.

marco-c: If the features didn't change with 0.5, it'd be nice to keep that.

marco-c:
You'll also need to run cargo update to update the Cargo.lock file

Cargo.toml
sylvestre: I tried but it doesn't seem that it is possible to do that anymore?!

marco-c: Ah it looks like the feature name changed from `deflate-zlib` to `deflate`.

calixteman: As far as I remember, there were some performance issues with the lastest zip lib.
So we need to be sure that we don't have a too big perf regression (if one of course).

sylvestre:
@marco-c @calixteman updated!

marco-c:
It looks like something must be changed for the new version:
```
UnsupportedArchive("Compression method not supported")
```

marco-c:
Closing in favor of #391.

PR #243, opened 2019-01-15
Never skip commits as it can cause problems with deployments

marco-c:


marco-c:
@calixteman this is why the deployment was not done.
The build corresponding to the tag is https://travis-ci.org/mozilla/grcov/jobs/479352880, which says:
```
$ git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '^appveyor.yml$' || {
    echo "Only appveyor.yml was updated, stopping build process."
    exit
  }
  
Only appveyor.yml was updated, stopping build process.
```

I think it's best to just avoid skipping commits, both for fixing this deployment problem and to avoid problems with code coverage uploads.

PR #255, opened 2019-02-17
Added -o flag to output to file

### LINKED ISSUE #208: Add -o option to set the output file
and -o - to output on stdout

byi8220:
Fixes #208.

Passes `cargo test` on multiple passes.

A notable change is that in this implementation the functions in `output.rs` no longer lock stdout for the entire duration from the start, instead writing to stdout normally without an explicit lock. I believe this is fine because the function that is called is the only thing that actually writes to output.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,6 @@ use std::fs::{self, File};
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::{env, process, thread};
-use std::option::Option;
 
 #[global_allocator]
 static GLOBAL: System = System;
@@ -65,7 +64,7 @@ fn main() {
     let mut path_mapping_file = "";
     let mut filter_option = None;
     let mut num_threads = num_cpus::get() * 2;
-    let mut output_file_path: Option<&str> = None;
+    let mut output_file_path = None;
 
     while i < args.len() {
         if args[i] == "-t" {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -64,6 +64,8 @@ fn main() {
     let mut path_mapping_file = "";
     let mut filter_option = None;
     let mut num_threads = num_cpus::get() * 2;
+    let mut output_file_path = "-";
```
marco-c: I'd make this an option, None by default and in output.rs then you can use stdio when this is None.

byi8220: I made it an option (None by default) in the newest commit.

src/main.rs
```diff
@@ -191,6 +194,16 @@ fn main() {
                 .parse()
                 .expect("Number of threads should be a number");
             i += 1;
+        } else if args[i] == "-o"{
+            if args.len() <= i + 1{
+                eprintln!("[ERROR]: Output file not specified.\n");
+                print_usage(&args[0]);
+                process::exit(1);
+            }
+            if &args[i + 1] != "-"{
```
marco-c: You can now remove this check

src/main.rs
```diff
@@ -64,6 +65,8 @@ fn main() {
     let mut path_mapping_file = "";
     let mut filter_option = None;
     let mut num_threads = num_cpus::get() * 2;
+    let mut output_file_path: Option<&str> = None;
```
marco-c: Nit: no need to specify the type, this way you can also avoid the `use Option`

src/main.rs
byi8220: So to confirm, I shouldn't check for the special case "-o -" to write to stdout?

PR #264, opened 2019-03-06
Map partial paths only for Java files

### LINKED ISSUE #262: Wrong coverage when multiple same file names
To reproduce:
 - git clone https://github.com/adrian17/grcov_test
 - CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Zno-landing-pads" cargo test
 - grcov target -s src/ -t lcov --llvm --branch --ignore-not-existing --ignore-dir "/*"  > lcov.info
 - lcov.info contains:
   DA:333,1
   DA:334,1
   DA:337,0
   DA:339,0
   DA:341,1
 for src/lib.rs which is obviously wrong.

It appears that we've two lib.rs files in gcnos:
 - src/lib.rs
 - /rustc/0ea22717a1e01fa535534b85a5347a7e49fc79de/src/libtest/lib.rs (on my system)

and because of path_rewriting stuff, these two files are finally the same !!

calixteman:
Fixes #262.


tests/test.rs
```diff
@@ -432,7 +432,7 @@ fn test_integration() {
         let entry = entry.unwrap();
         let path = entry.path();
 
-        if path.starts_with("tests/basic_zip_zip") || path.starts_with("tests/basic_zip_dir") {
+        if path.starts_with("tests/basic_zip_zip") || path.starts_with("tests/basic_zip_dir") || path.starts_with("tests/java") {
```
marco-c: Nit: you could put the file in test/java/test.java so you wouldn't have to skip the dir here

src/path_rewriting.rs
```diff
@@ -709,8 +725,8 @@ mod tests {
         for (abs_path, rel_path, result) in results {
             count += 1;
             assert!(abs_path.is_absolute());
-            assert!(abs_path.ends_with("tests\\class\\main.cpp"));
-            assert_eq!(rel_path, PathBuf::from("tests\\class\\main.cpp"));
+            assert!(abs_path.ends_with("tests\\java\\main.java"));
+            assert_eq!(rel_path, PathBuf::from("tests\\java\\main.java"));
             assert_eq!(result, empty_result!());
         }
         assert_eq!(count, 1);
```
marco-c: Could you also add a test where the partial path is mapped for a java file and not for a cpp or rust file?

PR #284, opened 2019-04-27
Replace MsQueue by channel::{Receiver, Sender} and update crossbeam to 0.7.1

### LINKED ISSUE #272: Update crossbeam
The latest version removed MsQueue, which we are using, so this isn't just a simple dependency update.

calixteman:
Fixes #272.

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -1,11 +1,18 @@
 #![feature(test)]
+extern crate crossbeam;
 extern crate grcov;
 extern crate rustc_hash;
 extern crate test;
 
+use crossbeam::crossbeam_channel::unbounded;
 use grcov::{CovResult, Function};
 use rustc_hash::FxHashMap;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+use std::thread;
 use test::{black_box, Bencher};
+
+use grcov::*;
 
 #[bench]
 fn bench_lib_merge_results(b: &mut Bencher) {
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -141,82 +141,92 @@ pub fn consumer(
...
-
-                                new_results.append(&mut try_parse!(
-                                    parse_gcov(&gcov_path),
-                                    work_item.name
-                                ));
-
-                                fs::remove_file(gcov_path).unwrap();
+    loop {
+        // If the queue is empty then it means that the producer
+        // doesn't push anything for the moment so it loops..
```
marco-c: This is not great, as we are spin-waiting here uselessly consuming resources since the pop is not blocking.

benches/lib.rs
```diff
@@ -70,3 +77,71 @@ fn bench_lib_merge_results(b: &mut Bencher) {
...
 }
+
+#[bench]
+fn bench_lib_consumer(b: &mut Bencher) {
+    let num_threads = 2;
+    let result_map: Arc<SyncCovResultMap> = Arc::new(Mutex::new(FxHashMap::with_capacity_and_hasher(20_000, Default::default())));
+    let (sender, receiver) = unbounded();
+    let source_root = None;
+    let working_dir = PathBuf::from("");
+    let gcno_buf: Vec<u8> = vec![
```
marco-c: You could read a gcno file, so you don't have to hardcode the bytes here

benches/lib.rs
```diff
@@ -70,3 +77,71 @@ fn bench_lib_merge_results(b: &mut Bencher) {
...
+        0, 236, 217, 93, 255, 2, 0, 0, 0, 109, 97, 105, 110, 0, 0, 0, 0, 2, 0, 0, 0, 102, 105,
+        108, 101, 46, 99, 0, 0, 1, 0, 0, 0, 0, 0, 65, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 67, 1, 3, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 67, 1, 3,
+        0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 69, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 69, 1, 8, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 102,
+        105, 108, 101, 46, 99, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0,
+    ];
+
+    b.iter(|| {
```
marco-c: This will benchmark thread creation too, but I hope it's an insignificant part of the runtime (given that you're sending a lot of messages).

benches/lib.rs
calixteman: I could. I copy/paste this gcno from a producer test, I think it's short enough to be inlined.

benches/lib.rs
calixteman: yeah I tried to remove this part from b(...) but to measure something we need to join in b and so we need to have parsers in b but parsers is not clonable because its content are not.

calixteman:
> Can you check if the perf is affected? (I wouldn't expect it to be, but just to be sure...)

I wrote a similar bench using the MSQueue and perfs are in the same order of magnitude.

src/lib.rs
```diff
@@ -205,12 +209,12 @@ pub fn consumer(
             ItemFormat::INFO | ItemFormat::JACOCO_XML => {
                 if let ItemType::Content(content) = work_item.item {
                     let buffer = BufReader::new(Cursor::new(content));
-                    if work_item.format == ItemFormat::INFO {
```
marco-c: There's something wrong here, maybe forgot to change back the indentation?

PR #286, opened 2019-05-07
Add an output format for FuzzManager

### LINKED ISSUE #253: Support coverage format used by FuzzManager
The server-side structure can be described by the following recursive definition:

pathObject
- int: linesTotal
- int: linesCovered
- int: linesMissed
- float: coveragePercent
- dict(name -> pathObject): children iff pathObject is a directory
- list(int): coverage iff pathObject is a file
- string: name

The coverage property is a list where each index represents the respective line in the file, starting with line 0 as the first line. If the value at a certain index is null, that means the line is not coverable. If the value is a positive integer or zero, then the value indicates how often the line has been covered.

Fields linesMissed and coveragePercent are redundant but we can save these calculations by just doing them once and storing the results, rather than having to make all of the calculations in the client on every display.

The name field is also redundant because it is identical to the key in the dict pointing to this pathObject, but it greatly helps the UI to have this property available.

calixteman:
Fixes #253.


src/main.rs
```diff
@@ -45,7 +45,7 @@ fn main() {
                                .long("output-type")
                                .value_name("OUTPUT TYPE")
                                .default_value("lcov")
-                               .possible_values(&["ade", "lcov", "coveralls", "coveralls+", "files"])
+                               .possible_values(&["ade", "lcov", "coveralls", "coveralls+", "files", "fuzzmanager"])
```
marco-c: Maybe we could find a nicer name for the format, something that says it's a recursive coverage format.

src/path_rewriting.rs
```diff
@@ -40,6 +40,16 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
     Ok(path)
 }
 
+fn normalize_path<P: AsRef<Path>>(path: P) -> PathBuf {
```
marco-c: Isn't there a Rust builtin method to do something like this?

src/main.rs
calixteman: Yes maybe... :)
I've absolutely no idea

src/fuzzmanager.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        Self {
+            total,
+            covered,
+            missed,
+            percent: Self::get_percent(covered, total),
+        }
+    }
+
+    pub fn add(&mut self, other: &Self) {
```
marco-c: I don't like that this method leaves the object in an unfinished state (as the percentage is not updated), but I see you did that to avoid recalculating the percentage all the time for dirs.
Maybe add a comment to say that after calling this method, you need to call set_percent too.

src/path_rewriting.rs
calixteman: There is canonicalize but you need to have a "true" path.
I checked the doc but didn't find anything except the components() trick.

marco-c: Yeah, we call canonicalize a few lines earlier, but it fails when the path is not there (which can be the case if you don't have a local clone of the repo).

src/main.rs
marco-c: recursive_coverage
recoverage
recover
recov

Do you like any of this?

La0: recov is nice.

We could also imply it's *our own* format: `mozcov`,  `mozilla_coverage`,

calixteman: or covcov :)

calixteman: +1 for mozcov

src/output.rs
```diff
@@ -107,28 +110,90 @@ pub fn output_activedata_etl(results: CovResultIter, output_file: Option<&str>)
...
+
+        let last_line = *result.lines.keys().last().unwrap_or(&0) as usize;
+        let mut lines: Vec<i64> = vec![-1; last_line];
+        for (line_num, line_count) in result.lines.iter() {
+            unsafe {
+                *lines.get_unchecked_mut((*line_num - 1) as usize) = *line_count as i64;
+            }
+        }
+        
+        prev_stats.borrow_mut().files.push(FMFileStats::new(path.file_name().unwrap().to_str().unwrap().to_string(), lines));
```
marco-c: Maybe it'd be better to pass result.lines to FMFileStats::new, so you don't have to iterate twice over them

src/main.rs
marco-c: I like any of recov, mozcov, covcov (hint to bugbug :P).

choller:
> LGTM, pending name change for the format, and verification of the output format from @choller (you can just look at test/test_fuzzmanager.json, https://github.com/mozilla/grcov/pull/286/files#diff-63e60beb8debf516244fe51cc5fb3c5d).

This looks correct to me. We will still have to make some changes on our side as well because we are pulling the revision out of the coveralls data right now because our format has no metadata fields (in our DB, having the metadata in the format again would be redundant, so we left it out). We can however specify it directly on the command line when calling our reporter tool.

Issue #281, opened 2019-04-20
Define alias for some variable types


marco-c:
As shown in #252, we are using some data types directly, which makes it hard when you want to switch to some other implementation of the same data type. E.g. in #252, we had to change a lot of places just to switch from HashMap to FxHashMap.

We should do something like https://github.com/mozilla/grcov/blob/14e4c76fd6951947c9406b404b715293a2104626/src/defs.rs#L51, defining aliases for specific usages so that when we want to try a different implementation of a data type we don't have to change all the code sites where the type is used.

lawrencek0:
Would it be alright if I work on this?

marco-c:
> Would it be alright if I work on this?

Sure, feel free!

PR #252, opened 2019-02-06
Evaluate using FxHashMap instead of HashMap

lawrencek0:
Fixes:  #235 

### Summary of Changes
* Added the rust-c crate for FxHashMap
* Replaced all uses of HashMap with FxHashMap

 `cargo build` works but I cannot seem to run the tests. I did try installing the LLVM packages but maybe I missed some.

PR #287, opened 2019-05-16
Implement alias for commonly used variable types

lawrencek0:
Fixes: #281 

### Summary of Changes
* Created and used alias for FxHashMap<String, Function>

I created this type alias because it was the one that was mostly changed during my PR (#252). I will look into the code base for commonly used types and will update the PR as I do so.

### DIFF
--- a/benches/filter.rs
+++ b/benches/filter.rs
@@ -3,13 +3,13 @@ extern crate grcov;
 extern crate rustc_hash;
 extern crate test;
 
-use grcov::{CovResult, Function};
+use grcov::{CovResult, Function, FunctionMap};
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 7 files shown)


lawrencek0:
I was hoping to alias more commonly used types. Should I still look into that @calixteman and make a different PR?/cc @marco-c

marco-c:
@lawrencek0 sure! It's better to split changes in multiple PRs rather than one big PR with all changes.

calixteman:
@lawrencek0 yep several little prs are better than a big one.
Anyway, in the future, if you plan to add something in PR, just add [WIP] in the PR title.

lawrencek0:
Got it!

PR #290, opened 2019-05-28
Normalize paths when canonicalization failed to do it

### LINKED ISSUE #289: Format option `covdir` is throwing backtrace
I wanted to try the new `covdir` option but I am always hitting this backtrace when trying to use it:

    # RUST_BACKTRACE=1 grcov /home/worker/firefox/ -t covdir -p /builds/worker/workspace/build/src > coverage.json                                                                                                                                                                        
    thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:347:21                                                                                                                                                                                            
    stack backtrace:                                                                                                                                                                                                                                                                                 
       0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace                                                                                                                                                                                                                                  
                 at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39                                                                                                                                                                                                                                
       1: std::sys_common::backtrace::_print                                                                                                                                                                                                                                                         
                 at src/libstd/sys_common/backtrace.rs:71                                                                                                                                                                                                                                            
       2: std::panicking::default_hook::{{closure}}                                                                                                                                                                                                                                                  
                 at src/libstd/sys_common/backtrace.rs:59                                                                                                                                                                                                                                            
                 at src/libstd/panicking.rs:197                                                                                                                                                                                                                                                      
       3: std::panicking::default_hook                                                                                                                                                                                                                                                               
                 at src/libstd/panicking.rs:211                                                                                                                                                                                                                                                      
       4: std::panicking::rust_panic_with_hook                                                                                                                                                                                                                                                       
                 at src/libstd/panicking.rs:474                                                                                                                                                                                                                                                      
       5: std::panicking::continue_panic_fmt                                                                                                                                                                                                                                                         
                 at src/libstd/panicking.rs:381                                                                                                                                                                                                                                                      
       6: rust_begin_unwind                                                                                                                                                                                                                                                                          
                 at src/libstd/panicking.rs:308                                                                                                                                                                                                                                                      
       7: core::panicking::panic_fmt                                                                                                                                                                                                                                                                 
                 at src/libcore/panicking.rs:85                                                                                                                                                                                                                                                      
       8: core::panicking::panic                                                                                                                                                                                                                                                                     
                 at src/libcore/panicking.rs:49                                                                                                                                                                                                                                                      
       9: grcov::output::output_covdir                                                                                                                                                                                                                                                               
      10: grcov::main                                                                                                                                                                                                                                                                                
      11: std::rt::lang_start::{{closure}}                                                                                                                                                                                                                                                           
      12: std::panicking::try::do_call                                                                                                                                                                                                                                                               
                 at src/libstd/rt.rs:49                                                                                                                                                                                                                                                              
                 at src/libstd/panicking.rs:293                                                                                                                                                                                                                                                      
      13: __rust_maybe_catch_panic                                                                                                                                                                                                                                                                   
                 at src/libpanic_unwind/lib.rs:85                                                                                                                                                                                                                                                    
      14: std::rt::lang_start_internal                                                                                                                                                                                                                                                               
                 at src/libstd/panicking.rs:272                                                                                                                                                                                                                                                      
                 at src/libstd/panic.rs:388                                                                                                                                                                                                                                                          
                 at src/libstd/rt.rs:48                                                                                                                                                                                                                                                              
      15: main                                 
      16: __libc_start_main           
      17: <unknown>

calixteman:
Fixes #289.

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -291,36 +291,39 @@ pub fn rewrite_paths(
         };
 
         // Get absolute path to the source file.
-        if let Some((abs_path, rel_path)) = get_abs_path(&source_dir, rel_path, &mut cache) {
-            if to_ignore_globset.is_match(&rel_path) {
-                return None;
-            }
-
-            if ignore_not_existing && !abs_path.exists() {
-                return None;
-            }
-
-            // Always return results with '/'.
-            let rel_path = PathBuf::from(rel_path.to_str().unwrap().replace("\\", "/"));
-
-            match filter_option {
-                Some(true) => {
-                    if !is_covered(&result) {
-                        return None;
-                    }
-                }
-                Some(false) => {
-                    if is_covered(&result) {
-                        return None;
-                    }
-                }
-                None => (),
-            };
-
-            Some((abs_path, rel_path, result))
-        } else {
-            None
+        let paths = get_abs_path(&source_dir, rel_path, &mut cache);
+        if paths.is_none() {
+            return None;
         }
+
+        let (abs_path, rel_path) = paths.unwrap();
+
+        if to_ignore_globset.is_match(&rel_path) {
+            return None;
+        }
+
+        if ignore_not_existing && !abs_path.exists() {
+            return None;
+        }
+
+        // Always return results with '/'.
+        let rel_path = PathBuf::from(rel_path.to_str().unwrap().replace("\\", "/"));
+
+        match filter_option {
+            Some(true) => {
+                if !is_covered(&result) {
+                    return None;
+                }
+            }
+            Some(false) => {
+                if is_covered(&result) {
+                    return None;
+                }
+            }
+            None => (),
+        };
+
+        Some((abs_path, rel_path, result))
     }))
 }
 
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -41,7 +41,7 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
 }
 
 
-pub fn normalize_path<P: AsRef<Path>>(path: P) -> PathBuf {
+pub fn normalize_path<P: AsRef<Path>>(path: P) -> Option<PathBuf> {
     // Copied from Cargo sources: https://github.com/rust-lang/cargo/blob/master/src/cargo/util/paths.rs#L65
     let mut components = path.as_ref().components().peekable();
     let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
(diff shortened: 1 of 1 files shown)


calixteman:
Not sure that's the correct way to deal with such path.
If relative path is "../foo/bar.cpp", then with this patch it'll become "foo/bar.cpp". I wonder if it's correct or not...

marco-c:
>  If relative path is "../foo/bar.cpp", then with this patch it'll become "foo/bar.cpp". I wonder if it's correct or not...

We can add a warning to let the user know that there might be something wrong.

src/path_rewriting.rs
```diff
@@ -134,6 +153,7 @@ fn get_abs_path(
     
     // Normalize the path in removing './' or '//'
```
marco-c: Add and '..'

src/path_rewriting.rs
```diff
@@ -40,14 +40,36 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
...
+        match component {
+            Component::Prefix(..) => unreachable!(),
+            Component::RootDir => {
+                ret.push(component.as_os_str());
+            }
+            Component::CurDir => {}
+            Component::ParentDir => {
+                if !ret.pop() {
+                    eprintln!("Warning: {:?} cannot be normalized because of \"..\", so skip it.", path.as_ref());
+                    return None;
```
marco-c: Maybe we should still normalize it ('../foo/bar' -> 'foo/bar'), but printing the warning so the user knows that something might be wrong.
This way at least we don't totally ignore a file, even if we might be attributing its coverage to a wrong file.

marco-c: If we ignore the path -> we don't risk attributing the coverage to a wrong file.
If we don't ignore the path -> we don't ignore the coverage of the file.

I'm not sure which one of the two is preferable...

calixteman: I think it's really a mistake to change "../foo/bar" into "foo/bar" because the two paths are different.
I think in such a case, the user should use source_dir option to be able to prepend the source_dir path to any too much relative paths... but right now source_dir needs to exist (see main.rs).

marco-c: OK, let's do 1 then.

src/path_rewriting.rs
```diff
@@ -264,34 +291,36 @@ pub fn rewrite_paths(
         };
 
         // Get absolute path to the source file.
-        let (abs_path, rel_path) = get_abs_path(&source_dir, rel_path, &mut cache);
-
-        if to_ignore_globset.is_match(&rel_path) {
-            return None;
-        }
+        if let Some((abs_path, rel_path)) = get_abs_path(&source_dir, rel_path, &mut cache) {
```
marco-c: Nit: You could check if None here and return early, so you don't have to incraese the indentation for all the rest.

PR #291, opened 2019-06-03
Guess path from gcno path for file with no parent dir

### LINKED ISSUE #288: Generated source files show up at toplevel
We recently noticed that several generated C/C++ source files show up as being on the top level, rather than the sub directory where they belong, e.g.:

    🐳  docker ~ # grcov /home/worker/firefox/ -t coveralls+ -p /builds/worker/workspace/build/src --token X --commit-sha Y > tmp.json && python -m json.tool tmp.json | grep bd_lf_apply_tmpl.c

    "name": "8bd_lf_apply_tmpl.c",

    🐳  docker ~ # find firefox/ | grep 8bd_lf_apply_tmpl
    firefox/media/libdav1d/8bd_lf_apply_tmpl.gcno
    firefox/media/libdav1d/8bd_lf_apply_tmpl.gcda

The expected behavior would be that `media/libdav1d/8bd_lf_apply_tmpl.c` is outputted for `name`. The respective `gcno/gcda` files are attached.

[8bd_lf_apply_tmpl.zip](https://github.com/mozilla/grcov/files/3224008/8bd_lf_apply_tmpl.zip)

calixteman:
Fix #288


src/lib.rs
```diff
@@ -186,7 +200,12 @@ pub fn consumer(
                             }
 
                             new_results
+                        };
+
+                        if guess_directory {
```
marco-c: Nit: could we call rename_single_files right before add_results, so we don't have to duplicate this code?

src/path_rewriting.rs
```diff
@@ -40,6 +40,18 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
...
+    if let Some(parent) = path.parent() {
+        if let Some(parent) = parent.to_str() {
+            parent == ""
+        } else {
+            false
+        }
+    } else {
+        false
+    }
+}
```
marco-c: Couldn't this be simplified to something like:
```Rust
PathBuf::from(path).parent() == Some(PathBuf::from(""))
```

src/path_rewriting.rs
```diff
@@ -40,6 +40,18 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
     Ok(path)
 }
 
+pub fn is_single_file(path: &str) -> bool {
```
marco-c: Could you find a nicer name for this?

src/lib.rs
calixteman: no, we can't because we need to have the stem.

PR #292, opened 2019-06-05
Update README.md to use new version in the example

kenkoooo:
When I copied & pasted the example to use grcov in my project, it failed in TravisCI, because it was `v0.4.1` and too old to use. To fix that, I updated README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Here is an example of .travis.yml file
 language: rust
 
 before_install:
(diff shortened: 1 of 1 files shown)


marco-c:
I wonder if there is a way to always point to the latest, it's going to be a pain to keep it up to date otherwise.

marco-c:
Here it is https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2, courtesy of @calixteman.

kenkoooo:
Thanks for your suggestion. I fixed it.

calixteman:
Thx for the patch

Issue #82, opened 2018-04-07
Write benchmarks for the output module


marco-c:
We should write benchmarks for the `output_*` functions in src/output.rs.

PR #294, opened 2019-06-15
Chore/output bench

TomPridham:
i added some benchmarks for the three output functions that i felt were benchmarkable. there weren't tests for coveralls, so i figured it was prolly safe to not bench it and the `output_files` function isn't doing a whole lot.

im not super sure if there is a better way to make the `CovResultIter` for the benches or if it should include more data or not. i tried messing around with cleaning it up, but couldn't ever get anything working because it gets moved into the closure and any variables declared outside of that box aren't guaranteed to live.

i'm also unsure if writing and removing the files like that is a good idea. it will overwrite and remove a file of the same name if it's there. `output_covdir` at least will write all of that data to the terminal if you pass it a `None`, tho, so we need to pass something in.

the function signature of the `thread::consumer` also changed and required an extra boolean arg in `benches/lib.rs` on line 118. all the other changes are white space from running rustfmt 

closes: https://github.com/mozilla/grcov/issues/82

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -81,19 +81,20 @@ fn bench_lib_merge_results(b: &mut Bencher) {
 #[bench]
 fn bench_lib_consumer(b: &mut Bencher) {
     let num_threads = 2;
-    let result_map: Arc<SyncCovResultMap> = Arc::new(Mutex::new(FxHashMap::with_capacity_and_hasher(20_000, Default::default())));
+    let result_map: Arc<SyncCovResultMap> = Arc::new(Mutex::new(
+        FxHashMap::with_capacity_and_hasher(20_000, Default::default()),
+    ));
     let (sender, receiver) = unbounded();
     let source_root = None;
     let working_dir = PathBuf::from("");
     let gcno_buf: Vec<u8> = vec![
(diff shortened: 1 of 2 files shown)


benches/lib.rs
```diff
@@ -81,19 +81,20 @@ fn bench_lib_merge_results(b: &mut Bencher) {
 #[bench]
 fn bench_lib_consumer(b: &mut Bencher) {
     let num_threads = 2;
-    let result_map: Arc<SyncCovResultMap> = Arc::new(Mutex::new(FxHashMap::with_capacity_and_hasher(20_000, Default::default())));
+    let result_map: Arc<SyncCovResultMap> = Arc::new(Mutex::new(
```
calixteman: this cosmetic change is unrelated to your patch, so just remove it.

benches/output.rs
```diff
@@ -0,0 +1,74 @@
...
+                )
+            }),
+    )
+}
+#[bench]
+fn bench_output_activedata_etl(b: &mut Bencher) {
+    b.iter(|| {
+        black_box(output_activedata_etl(
+            generate_cov_result_iter(),
+            Some("./temp"),
```
calixteman: use a file in a temporary dir (see tempdir crate)

calixteman:
@TomPridham thanks for the patch.

Issue #195, opened 2018-10-08
Use if-let chains when available to clean up some code


marco-c:
https://github.com/rust-lang/rust/issues/53667

crawfordleeds:
Hey @marco-c, curious about any performance or stylistic improvements for doing this. My understanding of if-let's are primarily for handling a single match pattern and ignoring the rest. As I look through the codebase, most comparisons seem to have two match patterns, e.g. in lib.rs:

```rust
match result.lines.entry(line_no) {
       btree_map::Entry::Occupied(c) => {
           *c.into_mut() += execution_count;
       }
       btree_map::Entry::Vacant(v) => {
           v.insert(execution_count);
       }
};
```

I could change that to and if-let like so:

```rust
if let btree_map::Entry::Occupied(c) = result.lines.entry(line_no) {
       *c.into_mut() += execution_count;
}
if let btree_map::Entry::Vacant(v) = result.lines.entry(line_no) {
       v.insert(execution_count);
}
```

Is this just the preferred syntax now, or are there performance improvements? Happy to start contributing to this project with this (feel free to assign to me), just wondering to get feedback on any technical improvements I don't know about or if this was just stylistic preference.

Cheers!

marco-c:
It's just a stylistic preference, but not to apply in cases like the one you mentioned.

A good example is https://github.com/mozilla/grcov/blob/5d6715d3779ec43a21f1568fc945c31684b2831f/src/gcov.rs#L7, where:
```
    match env::var("GCOV") {
        Ok(s) => s,
        Err(_) => "gcov".to_string(),
    }
```
would become:
```
if let Ok(s) = env::var("GCOV") {
  s
} else {
  "gcov".to_string()
}
```

crawfordleeds:
Great, thanks for the feedback. I'll take a shot at this and open a PR shortly. Thanks!

PR #295, opened 2019-07-15
Use if-let chains

crawfordleeds:
Per Issue #195, I have gone through and added some if-let chains instead of pattern matching. Would appreciate any feedback you may have. Thanks!

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -320,83 +320,157 @@ fn gcno_gcda_producer(
 
     for (gcno_stem, gcno_archive) in gcno_stem_archives {
         let stem = &gcno_stem.stem;
-        match gcda_stem_archives.get(stem) {
-            Some(gcda_archives) => {
-                let gcno_archive = *gcno_archive;
-                let gcno = format!("{}.gcno", stem).to_string();
-                let physical_gcno_path = tmp_dir.join(format!("{}_{}.gcno", stem, 1));
-                if gcno_stem.llvm {
-                    let mut gcno_buffer: Vec<u8> = Vec::new();
-                    let mut gcda_buffers: Vec<Vec<u8>> = Vec::with_capacity(gcda_archives.len());
-                    gcno_archive.read_in_buffer(&gcno, &mut gcno_buffer);
-                    for gcda_archive in gcda_archives {
-                        let mut gcda_buf: Vec<u8> = Vec::new();
-                        let gcda = format!("{}.gcda", stem).to_string();
-                        if gcda_archive.read_in_buffer(&gcda, &mut gcda_buf) {
-                            gcda_buffers.push(gcda_buf);
-                        }
+        if let Some(gcda_archives) = gcda_stem_archives.get(stem) {
+            let gcno_archive = *gcno_archive;
+            let gcno = format!("{}.gcno", stem).to_string();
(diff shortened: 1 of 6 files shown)


src/producer.rs
```diff
@@ -140,10 +140,13 @@ impl Archive {
     fn check_file(file: FilePath, checker: &Fn(&mut Read) -> bool) -> bool {
         match file {
             FilePath::File(reader) => checker(reader),
-            FilePath::Path(path) => match File::open(path) {
-                Ok(mut f) => checker(&mut f),
-                Err(_) => false,
-            },
+            FilePath::Path(path) => {
```
marco-c: This one in particular was probably a bit more readable with the match instead of if-let

src/producer.rs
```diff
@@ -250,23 +253,31 @@ impl Archive {
     pub fn extract(&self, name: &str, path: &PathBuf) -> bool {
         let dest_parent = path.parent().unwrap();
         if !dest_parent.exists() {
-            match fs::create_dir_all(dest_parent) {
-                Ok(_) => {}
-                Err(_) => panic!("Cannot create parent directory"),
-            };
+            if let Ok(_) = fs::create_dir_all(dest_parent) {}
```
marco-c: Maybe here you could do:
```Rust
if let Err(_) = fs::create_dir_all(dest_parent) {
  panic!("Cannot create parent directory")
}
```
Or even:
```Rust
fs::create_dir_all(dest_parent).expect("Cannot create parent directory");
```

src/producer.rs
```diff
@@ -250,23 +253,31 @@ impl Archive {
...
                 let mut zip = zip.borrow_mut();
                 let zipfile = zip.by_name(&name);
-                match zipfile {
-                    Ok(mut f) => {
-                        let mut file = File::create(&path).expect("Failed to create file");
-                        io::copy(&mut f, &mut file).expect("Failed to copy file from ZIP");
-                        true
-                    }
-                    Err(_) => false,
+                // match zipfile {
```
marco-c: Here you can remove the commented-out part.

marco-c:
Thanks, it looks good overall!

src/producer.rs
crawfordleeds: Oops, thanks :)

src/producer.rs
```diff
@@ -143,7 +143,7 @@ impl Archive {
             FilePath::Path(path) => match File::open(path) {
                 Ok(mut f) => checker(&mut f),
                 Err(_) => false,
-            },
+            }
```
marco-c: Nit: could you bring back the comma to avoid affecting 'blame' information

crawfordleeds: Of course, thanks for your patience.

marco-c: No worries, that's really minor!

crawfordleeds:
Any thoughts on why the codecov was so shot by these changes? I didn't really look into the tests, should I dig into that a bit?

marco-c:
> Any thoughts on why the codecov was so shot by these changes? I didn't really look into the tests, should I dig into that a bit?

No, no worries, I think it's just imprecise. We have to fix it somehow.

crawfordleeds:
> > Any thoughts on why the codecov was so shot by these changes? I didn't really look into the tests, should I dig into that a bit?
> 
> No, no worries, I think it's just imprecise. We have to fix it somehow.

Great, thanks!

PR #297, opened 2019-07-20
Add an html output

calixteman:


src/output.rs
```diff
@@ -353,6 +357,70 @@ pub fn output_files(results: CovResultIter, output_file: Option<&str>) {
...
+pub fn output_html(results: CovResultIter, output_dir: Option<&str>, num_threads: usize) {
+    let output = if let Some(output_dir) = output_dir {
+        PathBuf::from(output_dir)
+    } else {
+        PathBuf::from("./html")
+    };
+
+    if output.exists() {
+        if !output.is_dir() {
+            eprintln!("./html is not a directory");
```
marco-c: The error message should actually specify the real output_dir

src/output.rs
```diff
@@ -353,6 +357,70 @@ pub fn output_files(results: CovResultIter, output_file: Option<&str>) {
...
+            return;
+        }
+    } else if std::fs::create_dir(&output).is_err() {
+        eprintln!("Cannot create directory ./html");
```
marco-c: Here too

Issue #300, opened 2019-08-01
Allow to set branch name for coveralls report


svartalf:
Branch name is hardcoded to `"master"` right now, it would be nice to be able to change it via CLI argument

https://github.com/mozilla/grcov/blob/6471bb938c1306be97b830ae72571cb2e1ef0094/src/output.rs#L334-L339

PR #301, opened 2019-08-04
Allow user input to set branch name in coveralls report

crawfordleeds:
This PR is intended to implement the request in issue #300. It should maintain all existing functionality by default to set the "master" branch unless additional input is provided by the user.

I ran `cargo fmt`, which is why there are a few stylistic changes. Sorry, I didn't realize the rest of the codebase wasn't run with `cargo fmt`. If that's really distracting I can revert and just add the functional changes with no stylistic changes.

Otherwise, I would greatly appreciate any feedback if this makes sense or if there's anything else needed to merge. Thanks!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,7 +147,12 @@ fn main() {
 
                           .arg(Arg::with_name("guess_directory")
                                .long("guess-directory-when-missing"))
-
+                          
+                          .arg(Arg::with_name("vcs_branch")
+                               .help("Set the branch for coveralls report. Defaults to 'master'")
+                               .long("vcs-branch")
+                               .value_name("VCS BRANCH")
+                               .takes_value(true))
                           .get_matches();
 
     let paths: Vec<_> = matches.values_of("paths").unwrap().collect();
@@ -179,7 +184,7 @@ fn main() {
     let service_name = matches.value_of("service_name").unwrap_or("");
     let service_number = matches.value_of("service_number").unwrap_or("");
     let service_job_number = matches.value_of("service_job_number").unwrap_or("");
-    let branch = matches.value_of("branch").unwrap_or("master");
+    let branch = matches.value_of("vcs_branch").unwrap_or("master");
     let num_threads: usize = matches
         .value_of("threads")
         .unwrap()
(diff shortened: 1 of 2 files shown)


marco-c:
> I ran cargo fmt, which is why there are a few stylistic changes. Sorry, I didn't realize the rest of the codebase wasn't run with cargo fmt. If that's really distracting I can revert and just add the functional changes with no stylistic changes.

Yes, please, it'll make it easier to review.
We'll run the codebase through fmt later (and setup pre-commit to do it automatically so we always merge "styled" code).

crawfordleeds:
Great, just pushed the updates! I reverted because it looks like you can't cmd+z out of the formatOnSave feature in VSCode.

Unless you're already addressing it, I'm happy to make the formatting/pre-commit hook PR for review.

src/main.rs
```diff
@@ -179,6 +179,7 @@ fn main() {
     let service_name = matches.value_of("service_name").unwrap_or("");
     let service_number = matches.value_of("service_number").unwrap_or("");
     let service_job_number = matches.value_of("service_job_number").unwrap_or("");
+    let branch = matches.value_of("branch").unwrap_or("master");
```
marco-c: You'll need to add the option to the clap App definition above too

marco-c: Ah, I forgot we already have a `branch` option, but it is to enable branch coverage parsing!

So we need to change the name of this option, you can call it vcs_branch

crawfordleeds: oh, my bad. Give me 5 :)

crawfordleeds: Just pushed the update. Not sure if you have a preference on the variable name too but I changed the var name to vcs_branch, not just the input string.

src/main.rs
```diff
@@ -179,6 +185,7 @@ fn main() {
     let service_name = matches.value_of("service_name").unwrap_or("");
     let service_number = matches.value_of("service_number").unwrap_or("");
     let service_job_number = matches.value_of("service_job_number").unwrap_or("");
+    let vcs_branch = matches.value_of("vcs_branch").unwrap_or("master");
```
marco-c: Nit: you can define the default value using [default_value](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.default_value).

crawfordleeds: Much better, thanks for the feedback.

Issue #306, opened 2019-08-08
Add `cargo clippy` pre-commit hook


ghost:
Per pull request #305, we have added pre-commit hooks for formatting and testing. Clippy was included but commented out due to a large number of errors and warnings that require resolution with clippy. The consensus seems to be that we should resolve those errors/warnings and include clippy.

marco-c:
Fixed by #309.

crawfordleeds:
Hey @marco-c, I'm afraid I may have misled you on PR #309. There was one remaining issue with running clippy. I tried to add it as part of a review but that probably wasn't clear enough.

With the current state of the repo, if you run `cargo clippy --all-targets --all -- -D warnings` one error lingers. This makes it such that the pre-commit hook will fail so people will have to commit with the `--no-verify` flag until this is resolved.

```
error: trait objects without an explicit `dyn` are deprecated
   --> src/producer.rs:141:45
    |
141 |     fn check_file(file: FilePath, checker: &Fn(&mut dyn Read) -> bool) -> bool {
    |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn Fn(&mut dyn Read) -> bool`
    |
    = note: `-D bare-trait-objects` implied by `-D warnings`

error: trait objects without an explicit `dyn` are deprecated
   --> src/producer.rs:141:45
    |
141 |     fn check_file(file: FilePath, checker: &Fn(&mut dyn Read) -> bool) -> bool {
    |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn Fn(&mut dyn Read) -> bool`
    |
    = note: `-D bare-trait-objects` implied by `-D warnings`

error: aborting due to previous error

error: Could not compile `grcov`.
warning: build failed, waiting for other jobs to finish...
error: aborting due to previous error

error: Could not compile `grcov`.
```

When I try to just add the `dyn` keyword to the argument, per the help message, I see a whole new can of errors. Do you have any recommendations on how to address that? Thanks!

marco-c:
Oh, ok. I've committed f0ed4cdcdcdfda9132bbe0be60200d22e922aeec to temporarily disable Clippy (so that other contributors won't be confused if pre-commit fails).

I think the "dyn" error is just hiding the others. When you add "dyn", Clippy will show the other errors too, which have to be fixed one at a time.

marco-c:
[...]

PR #309, opened 2019-08-09
Add Clippy fixes and add to pre-commit

crawfordleeds:
Per issue #306, this PR solves all linting errors and warnings from running `cargo clippy`. Additionally, it adds clippy to the pre-commit hook.

I haven't yet completed all required changes, but expect to be able to finish up sometime in the next few days. Feel free to let me know if you have any feedback so far. Thanks!


src/producer.rs
```diff
@@ -36,13 +36,14 @@ pub struct GCNOStem {
...
-
-    #[cfg(windows)]
-    let path = path.replace("\\", "/");
+    path.to_str().unwrap().to_string()
+}
 
-    path
+#[cfg(windows)]
+fn clean_path(path: &PathBuf) -> String {
+    path.replace("\\", "/")
```
marco-c: Aargh, it looks like this has to be converted to a string first (see the failure on Appveyor)

crawfordleeds: Ah, thanks! Just pushed a fix.

Issue #308, opened 2019-08-09
Integration Tests Failing


crawfordleeds:
I'm currently having issues getting all the integration tests to pass. I've had this issue since I first cloned the repository which I have mostly ignored. However, it is now causing problems now that the pre-commit hooks are enabled on this repository. Since it doesn't pass the tests I have to commit with --no-verify.

I was hoping to get some help either identifying the issue on my machine or determining a path forward to fix if this is an issue with the project (seems unlikely since all CI tools pass the tests).

Here's the last of the output when I run `cargo test`:

```
thread 'test_integration' panicked at 'assertion failed: `(left == right)`
  left: `Array([Null, Null, Null, Null, Number(1), Null, Number(4), Number(1), Null, Number(1), Number(2)])`,
 right: `Array([Null, Null, Null, Null, Number(1), Null, Number(1), Number(1), Null, Number(1), Number(1)])`: Got correct coverage for "main.cpp"', tests/test.rs:302:13
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    test_integration

test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test test'
```

I've tried this with a few different versions of rustc including 1.32.0 nightly, 1.36 stable, and 1.38.0 nightly all with the same result.

Does anyone have any thoughts on what the issue may be? Thanks for the help!

marco-c:
What version of LLVM and GCC do you have?

BTW, since integration tests are hard to set up at the moment, we should make pre-commit not run them.

calixteman:
FYI, the way to run all the tests:
GCOV=gcov-7 CLANG_CXX=clang++-9 GCC_CXX=g++-7 cargo test

The integration tests are failing with a g++ version upper than 8 (included).

crawfordleeds:
Thank you both for the feedback.

@marco-c, I'm on macOS and running Apple LLVM version 10.0.1 (clang-1001.0.46.4). I'm actually not sure what version of LLVM that corresponds to - I think they stopped sharing that in Apple LLVM version 7. I did some searching and [this table](https://en.wikipedia.org/wiki/Xcode#Toolchain_versions) seems to imply LLVM version 7.0.0.

I've never done it, but do you recommend installing LLVM directly either from source or through homebrew?

I'll work on removing the integration tests from pre-commit. That'll be helpful for me.

[...]

PR #310, opened 2019-08-11
Run Only Unit Tests in Pre-Commit

crawfordleeds:
Per the conversation in https://github.com/mozilla/grcov/issues/308#issuecomment-519827076, the pre-commit hook should only run unit tests.

This PR changes the pre-commit configuration to only run unit tests and documents the option in the README for reference.


test/rust/main.rs
```diff
@@ -0,0 +1 @@
+
```
marco-c: Why are this file and test/rust/src/main.rs necessary?

crawfordleeds: I'm not sure that they are necessary. They currently exist in the master branch so the existence of the files is a carryover of previous work but it looks like for some reason I added a couple blank lines. I don't recall actually modifying these files manually so I'm guessing the line was added by my IDE?

I just tried deleting these two files and the tests still pass.

I'll remove the blank lines and commit. Think we should remove these files in a new PR?

marco-c: We can leave them, I thought you were adding them now (the GitHub interface confused me).

crawfordleeds: Cool - I was confused at first too. I'm still somewhat confused how those changes ended up there in the first place 🙃

PR #313, opened 2019-09-04
[WIP]: Complete clippy warning fixes

crawfordleeds:
Per #306, this PR includes the last resolutions for all clippy warnings on the existing codebase.

It is a WIP, more details coming soon.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -98,7 +98,7 @@ macro_rules! read_u_le {
         $buf.pos += size;
         if $buf.pos <= $buf.buffer.len() {
             Ok(unsafe {
-                *std::mem::transmute::<*const u8, *const $ty>($buf.buffer[start..].as_ptr())
+                *$buf.buffer[start..].as_ptr() as $ty
             }
             .to_le())
         } else {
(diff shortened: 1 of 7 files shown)


src/reader.rs
```diff
@@ -98,7 +98,8 @@ macro_rules! read_u_le {
         $buf.pos += size;
         if $buf.pos <= $buf.buffer.len() {
             Ok(unsafe {
-                *$buf.buffer[start..].as_ptr() as $ty
+                *std::mem::transmute::<*const u8, *const $ty>($buf.buffer[start..].as_ptr())
+                // *$buf.buffer[start..].as_ptr() as $ty
```
calixteman: With your code, you're just converting an u8 to $ty.
Since the data are are correctly aligned in the gcno/gcda files, it's probably the fastest way to get the $ty value.

crawfordleeds: Hey @calixteman, I'm not very familiar with transmute, is your description of my change not what's happening with transmute?

Using what I wrote, `*$buf.buffer[start..].as_ptr() as $ty`, clippy passes but the tests do not. That's why I reverted to `*std::mem::transmute::<*const u8, *const $ty>($buf.buffer[start..].as_ptr())` and added `#[allow(clippy::transmute_ptr_to_ptr)]`.

Could you guide me in the right direction to solve the [transmute_ptr_to_ptr lint](https://rust-lang.github.io/rust-clippy/current/index.html)?

Thanks!

svenfoo:
With #625 merged, this PR seems obsolete and can probably be closed.

Issue #226, opened 2018-10-30
Prettify all JSON expected files under tests/


marco-c:
Instead of https://github.com/mozilla/grcov/blob/aee0b46ebcf4e4f37482b4e3d6acb8cf832c63fa/tests/include2/expected_llvm_7_linux.coveralls, they should all be like https://github.com/mozilla/grcov/blob/71c71beebee5ad62329aaea2d2df88de366bbe5d/tests/basic/expected_llvm.coveralls.

arshadkazmi42:
@marco-c Is it up for grab? I can take this up

marco-c:
@harshagarwal7272 started on this in #228, but I'm not sure if they intend on continuing the work. @harshagarwal7272 do you want to keep working on this?

Mika-net:
@marco-c if it's alright I will start on this #228 and submit a PR with it either today or tomorrow.

marco-c:
Sure, feel free, that PR was abandoned.

Mika-net:
@marco-c I've submitted a PR, lmk if anything looks off.

marco-c:
The ADE files (*.ade) should not be prettified (the tests are reading one line at a time: https://github.com/mozilla/grcov/blob/6bc906f1065e51ebb78c1a27fbe8535228ebba39/tests/test.rs#L221-L231). Sorry, I forgot to mention that :(

Mika-net:
That would explain that, no worries. I'll try with just the coveralls

marco-c:
> I'll try with just the coveralls

covdir should be OK to prettify too

Issue #258, opened 2019-02-19
Create a Docker image to run tests (especially integration ones)


calixteman:

marco-c:
Good idea!

crawfordleeds:
I was just thinking of making this same issue when I saw this. I'm slammed tomorrow but will try and get started on this on Wednesday.

PR #315, opened 2019-09-06
prettified all ugly json under /tests/ per #226

Mika-net:
Prettified all ugly JSON all non integration tests pass numbering 83.


marco-c:
Tests are failing, so there's likely something wrong in one or more of the files you prettified.

Mika-net:
@marco-c I rolled back the changes locally but still have the same failing test, I'm looking into how to fix it but It looks like the issue is that the expected array isn't matching the result, to test I'm altering the expected to match the actual. This error is in a file that is already formatted prior to this pr. Let me know if you don't have the same issue.

marco-c:
@Mika-net the tests are passing on CI and on my machine. I know setting up the integration tests locally is not that easy unfortunately... @calixteman what were the versions of GCC and LLVM required for integration tests?

A (slow) workaround for you to find the "bad" files would be to push your changes a few at a time and let CI run the tests for you.

Mika-net:
@marco-c  thanks for the feedback! For the tests all I did was cargo build/cargo test. Was there more to the process? I'll start on the slow work around.

marco-c:
The integration tests don't work with all versions of GCC and LLVM, so the requirements to make them pass are pretty strict (see #308 and #258).

Mika-net:
I don't know if this will ping yall alot but I'll let ya know when I'm done isolating the issue via these automated tests

Mika-net:
@marco-c I've tested it with an extremely small change and it throws this, thread 'test_integration' panicked at  
``` called Result::unwrap()` on an `Err` value: Error("EOF while parsing an object", line: 1, column: 1)', src/libcore/result.rs:999:5 ```

Is it possible the integration tests cant parse these JSON files if they're prettified?

marco-c:
> Is it possible the integration tests cant parse these JSON files if they're prettified?

Answered at https://github.com/mozilla/grcov/issues/226#issuecomment-529973994 (the ADE files are not JSON files, but files where each line is a JSON object, so they can't be prettified).

Mika-net:
@marco-c sorry about all the back and forth, but it should  be good now.

marco-c:
Looks good so far, but I think you can also prettify the covdir files.

marco-c:
Also, I think you missed the files under tests/basic/ and tests/switch/.

Mika-net:
For test  basic and switch the only ones not already prettified are the .ade files, and all but one of the .covdir files was already prettified.

marco-c:
Oh, OK!

PR #327, opened 2019-10-07
Replace help output in readme with up-to-date version

badboy:
[Recent changes](https://github.com/mozilla/grcov/pull/319) renamed a parameter. This wasn't reflected in the readme's displayed help output.

I simply replaced the data in the readme with the output of `grcov --help`


marco-c:
Oops, maybe in the README we should suggest calling `grcov --help` to get the list of options, otherwise we'll always forget to update them.

PR #400, opened 2020-03-17
When paths are different in gcno/gcda archives then try to match file name + stamp

calixteman:
Normally we expect to have `a/b/c/foo.gcno` and `a/b/c/foo.gcda` in archives but for any reason the base paths can be different.
So the idea is to use filename + stamp (or checksum for llvm gcno/gcda) as a unique identifier when paths don't match.
In case where filename + stamp doesn't identify an unique stem then we fail.


aureooms:
Does this fix what I am experiencing here: [travis lists a bunch of gcna/gcdo files covering all executables and libraries](https://travis-ci.org/github/aureooms/memoize/builds/711966066#L742) in the zip archive but [codecov only lists the library source files](https://codecov.io/gh/aureooms/memoize/list/rewrite-rust/).

marco-c:
@calixteman do we still want to do this?

marco-c (changes requested):
If we still want to do this, the first step will be to rebase it.

PR #404, opened 2020-03-25
Use the correct amount of memory for the buffers

calixteman:
The amount of required memory is just what we need.
We've ~5% of performance improvements (because of no resize).

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -217,34 +217,44 @@ impl Archive {
         }
     }
 
-    pub fn read_in_buffer(&self, name: &str, buf: &mut Vec<u8>) -> bool {
+    pub fn read(&self, name: &str) -> Option<Vec<u8>> {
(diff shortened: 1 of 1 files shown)


src/producer.rs
```diff
@@ -217,34 +217,44 @@ impl Archive {
...
+                    Err(_) => None,
                 }
             }
-            ArchiveType::Dir(ref dir) => match File::open(dir.join(name)) {
-                Ok(mut f) => {
-                    f.read_to_end(buf).expect("Failed to read gcda file");
-                    true
+            ArchiveType::Dir(ref dir) => {
+                let path = dir.join(name);
+                let metadata = fs::metadata(&path).unwrap();
```
marco-c: Previously we were not failing when we were not able to open a file, we should probably keep that behavior

PR #408, opened 2020-03-26
Rewrite lcov parser

calixteman:
This new version give us a 20% speed improvement.
So it could help to reduce data retention in the queue especially for system with 1 or 2 threads.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -148,182 +148,165 @@ pub fn parse_lcov(
 
     let mut line = 0;
 
-    loop {
-        if let Some(c) = iter.next() {
-            line += 1;
-            match *c {
-                b'e' => {
-                    // we've a end_of_record
-                    results.push((
-                        cur_file.unwrap(),
-                        CovResult {
-                            lines: cur_lines,
-                            branches: cur_branches,
-                            functions: cur_functions,
-                        },
-                    ));
+    while let Some(c) = iter.next() {
+        line += 1;
+        match *c {
+            b'e' => {
+                // we've a end_of_record
+                results.push((
+                    cur_file.unwrap(),
+                    CovResult {
+                        lines: cur_lines,
+                        branches: cur_branches,
+                        functions: cur_functions,
+                    },
+                ));
 
-                    cur_file = None;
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -132,103 +132,198 @@ pub fn parse_lcov<T: Read>(
...
-                        let taken = try_next!(values, l) != "-";
-                        add_branch(&mut cur_branches, line_no, branch_number, taken);
+
+                    let key = iter
+                        .take_while(|&&c| c != b':')
+                        .fold(*c as u32, |r, &x| r * (1 << 8) + u32::from(x));
+                    match key {
+                        SF => {
+                            // SF:string
+                            cur_file = Some(iter.take_while(|&&c| c != b'\n' && c != b'\r').fold(
```
marco-c: Wouldn't `collect` work instead of `fold`? Is this way faster?

src/parser.rs
```diff
@@ -132,103 +132,198 @@ pub fn parse_lcov<T: Read>(
...
+                                .fold(0, |r, &x| r * 10 + u32::from(x - b'0'));
+                            if iter.peek().is_none() {
+                                return Err(ParserError::InvalidRecord(format!(
+                                    "DA at line {}",
+                                    line
+                                )));
+                            }
+                            let execution_count = if let Some(c) = iter.next() {
+                                if *c == b'-' {
+                                    iter.take_while(|&&c| c != b'\n').fold(0, |_, _| 0)
```
marco-c: Instead of `.fold(0, |_, _| 0)` to consume iterators, it might be clearer to use `count()` or `last()`

src/parser.rs
```diff
@@ -132,103 +132,198 @@ pub fn parse_lcov<T: Read>(
...
+                        }
+                        _ => {
+                            iter.take_while(|&&c| c != b'\n').fold(0, |_, _| 0);
+                        }
                     }
                 }
-                _ => {}
             }
+        } else {
+            break;
```
marco-c: Nit: add a comment `// End of input`, otherwise one has to go back to the beginning to understand what this break does. Or, even better, you could use `while let`

PR #409, opened 2020-03-29
Fix travis setting on README

kngwyu:
Now the released grcov binary needs tcmalloc, so I think it's helpful to recommend users to install it on travis.


marco-c:
Thanks for the PR, but we are going to drop the tcmalloc dependency from normal releases with #410.

PR #412, opened 2020-03-30
Update readme about tcmalloc

calixteman:


marco-c:
I'd say we can keep this as an internal feature (most likely we're the only ones benefiting from it) and so not document it to avoid confusion.

PR #416, opened 2020-04-02
Exclusion Patterns for Lines and Branches

### LINKED ISSUE #261: Support excluding some lines, functions, macros, branches from coverage report
This is a feature request.
lcov has special markers that can be used to exclude parts of code from coverage report. For example, LCOV_EXCL_LINE, LCOV_EXCL_BR_LINE and their START/STOP multiline variants.

It would be nice to have something like this here too.

For example, macros sometimes generate always-true branches that grcov will report as partially-covered. In this case, it would be nice to be able to explicitly exclude them (well, code generated by them) from coverage report.

jcdickinson:
Adds line and branch exclusion, following the names of the vars in lcovrc. The lines are removed during path rewriting (as the full path is resolved during this phase). Resolves #261

Dangling regions at the end of files are permitted for e.g. `mod tests {`

Also resolves some failing tests as well as compilation errors on Windows (see individual commits).


src/file_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+        let file = if let Ok(file) = file {
+            file
+        } else {
+            return Vec::new();
+        };
+
+        let mut ignore_br = false;
+        let mut ignore = false;
+
+        file.split("\n")
```
calixteman, line 78: you should add comments here to explain each step

src/main.rs
```diff
@@ -182,6 +182,42 @@ fn main() {
                                .default_value("stderr")
                                .value_name("LOG")
                                .takes_value(true))
+                               
+                          .arg(Arg::with_name("excl_line")
+                               .help("Lines containing this marker will be excluded.")
```
calixteman, line 187: nit: you should say where are these "Lines"

jcdickinson:
Thanks for all the great feedback!

> Would it be possible to multi-thread the filter part ?

I had a go at it using Rayon, adding inline comments in a few moments.

src/path_rewriting.rs
```diff
@@ -280,78 +272,76 @@ pub fn rewrite_paths(
         }
     }
 
-    let mut cache: Option<PathBuf> = None;
```
jcdickinson, line 282: Is removing this case correc? It seems as though it was tied to the in-order evaluation of files (which no longer applies because parallel is by nature unordered).

.travis.yml
```diff
@@ -126,7 +126,7 @@ script:
     - |
       if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then
         zip -0 ccov.zip `find . \( -name "grcov*.gc*" -o -name "llvmgcov.gc*" \) -print`;
-        ./target/debug/grcov ccov.zip -s . -t lcov --llvm --branch --ignore-not-existing --ignore "/*" -o lcov.info;
```
jcdickinson, line 129: NB: I have ignored `#[derive(` lines now that grcov supports it 😉

jcdickinson:
FYI: TravisCI doesn't seem to be reporting back on successful builds: https://travis-ci.org/github/mozilla/grcov/builds/670724791

test/java/skip.java
```diff
@@ -0,0 +1,20 @@
+class Test {
```
calixteman, line 1: Is this file associated to a test ?

jcdickinson: It is used by these two tests: https://github.com/mozilla/grcov/pull/416/files#diff-35517b643f4f9034bdad1c30fde730adR1361

calixteman: Sorry but I don't see any test in this patch. Or maybe I missed something...

jcdickinson: The tests are right at the end of the collapsed diff for src/path_rewriting.rs

README.md
```diff
@@ -38,6 +38,18 @@ FLAGS:
 
 OPTIONS:
         --commit-sha <COMMIT HASH>                   Sets the hash of the commit used to generate the code coverage data
+        --excl-br-line <regex>
+            Lines in covered files containing this marker will be excluded.
```
rivy, line 42: Help/docs about branch coverage exclusions should be probably be annotated differently, such as
... `Lines in covered files containing this marker will be excluded from branch coverage.` ?

README.md
```diff
@@ -38,6 +38,18 @@ FLAGS:
...
+        --excl-br-start <regex>
+            Marks the beginning of an excluded section. The current line is part of this section.
```
rivy, line 44: ... `Marks the beginning of a section which is excluded from branch coverage. The current line is part of this section.` ?

README.md
```diff
@@ -38,6 +38,18 @@ FLAGS:
...
+        --excl-br-stop <regex>
+            Marks the end of an excluded section. The current line is part of this section.
```
rivy, line 46: ... `Marks the end of a section excluded from branch coverage. The current line is part of this section.` ?

src/main.rs
```diff
@@ -183,6 +183,42 @@ fn main() {
...
+                                .takes_value(true))
+
+                            .arg(Arg::with_name("excl-stop")
+                                .help("Marks the end of an excluded section. The current line is part of this section.")
+                                .long("excl-stop")
+                                .value_name("regex")
+                                .takes_value(true))
+
+                          .arg(Arg::with_name("excl-br-line")
+                               .help("Lines in covered files containing this marker will be excluded.")
```
rivy, line 205: `.help("Lines in covered files containing this marker will be excluded from branch coverage.")` ?

src/main.rs
```diff
@@ -183,6 +183,42 @@ fn main() {
...
+                               .long("excl-br-line")
+                               .value_name("regex")
+                               .takes_value(true))
+
+                            .arg(Arg::with_name("excl-br-start")
+                                .help("Marks the beginning of an excluded section. The current line is part of this section.")
```
rivy, line 211: ... `.help("Marks the beginning of a section which is excluded from branch coverage. The current line is part of this section.")` ?

src/main.rs
```diff
@@ -183,6 +183,42 @@ fn main() {
...
+                                .long("excl-br-start")
+                                .value_name("regex")
+                                .takes_value(true))
+
+                            .arg(Arg::with_name("excl-br-stop")
+                                .help("Marks the end of an excluded section. The current line is part of this section.")
```
rivy, line 217: ... `.help("Marks the end of a section which is excluded from branch coverage. The current line not part of this section.")`

rivy:
I'm new to `gcov`, `grcov`, `gcovr`, `lcov`, ...

Could you suggest how this would best be used?

For example, what would be a good method of ignoring `assert_eq!(...)` incomplete branch coverage?

It looks like the `lcov` method (used fairly widely) is to use `LCOV_EXCL_LINE` somewhere on the excluded line. It doesn't look like the tools require any extra option flags.

This patch looks like it's generic without suggesting a similar default implementation.

ref: <https://stackoverflow.com/questions/3555083/how-do-i-tell-gcov-to-ignore-un-hittable-lines-of-c-code>
ref: <https://manpages.debian.org/jessie/lcov/geninfo.1.en.html>

And, thanks for the patch!

jcdickinson:
> And, thanks for the patch!

Always a pleasure!

> For example, what would be a good method of ignoring `assert_eq!(...)` incomplete branch coverage?

In some projects, tests are contained in different directories entirely - those projects already have usable ignores in grcov (just ignore the entire file with the existing file exclusion patterns). The other approach that seems widespread in Rust is to have a `mod tests` at the bottom of the file. In that scenario, you could ignore every line in your tests by passing `--excl-br-start "mod tests \{" --excl-start "mod tests \{"`. That would begin a range in each file starting at `mod tests {` and only terminating at the end of the file (because there is no associated `-stop` argument).

In addition, the derive Rust attribute (e.g. `#[derive(Debug, Eq, PartialEq)]`) can usually be assumed to be working code, because there are tests in the Rust stdlib that make sure that this derive macro works. This is where the per-line options are relevant `--excl-br-line "#\[derive\(" --excl-line "#\[derive\("`. This fails to account for lines with good macro hygiene (i.e. fields in structs are still counted as uncovered), but it's a start.

[Here's a coverage report demonstrating both of the above](https://coveralls.io/builds/30002258/source?filename=racemus-binary/src/nbt/value.rs), and the [Github workflow configuration file](https://github.com/jcdickinson/racemus/blob/master/.github/actions-rs/grcov.yml#L10) that contains the regex patterns.

I probably wouldn't ignore assert_eq! (as that may be used in non-test code), but you could ignore it by passing `--excl-br-line "assert_eq!" --excl-line "assert_eq!"`.

Having two parameters is a bit redundant, but I assumed there is a good reason why lcov differentiates between ignored lines and branches.

> It looks like the `lcov` method (used fairly widely) is to use `LCOV_EXCL_LINE` somewhere on the excluded line. It doesn't look like the tools otherwise require any extra option flags.

lcov allows you to customize this with a machine-wide [`lcovrc` file](http://ltp.sourceforge.net/coverage/lcov/lcovrc.5.php) - which isn't a great solution in my opinion (as all developer machines have to either have no `lcovrc` file, or have to have an `lcovrc` file that agrees for all projects that they work on). It seems to be a better solution to pass these on the CLI so that each Git/whatever repo can decide how best to approach ignores (and use intelligent ignores for idiomatic patterns in the language that they are using, such as my above examples). The names of the CLI args closely follow the names of the values in `lcovrc`.

If you wanted to use this ignore pattern, you could pass `--excl-line LCOV_EXCL_LINE` to grcov.

> This patch looks like it's generic without suggesting a similar default implementation. Would including the `lcov` "standard" keywords by default be a good idea?

`lcovrc` does default to what you indicated, but that means you have ignore patterns whether you want them or not. This smells like an anti-feature to me, but I'm happy to concede to the principle of least surprise and go with what lcov does.

I'll rectify the remainder of your feedback tomorrow. Thanks!

rivy:
> I probably wouldn't ignore assert_eq! (as that may be used in non-test code), but you could ignore it by passing `--excl-br-line "assert_eq!" --excl-line "assert_eq!"`.
 
So, from some google research and your feedback, it looks like I should just use something like `--excl-br-line "assert(_eq|_ne)?!"` when calling `grcov`. That would still count the line(s) for coverage but ignore that the internal branch is not taken (avoiding coverage warnings).

Sound correct?

> In addition, the derive Rust attribute (e.g. `#[derive(Debug, Eq, PartialEq)]`) can usually be assumed to be working code, because there are tests in the Rust stdlib that make sure that this derive macro works. This is where the per-line options are relevant `--excl-br-line "#\[derive\(" --excl-line "#\[derive\("`. This fails to account for lines with good macro hygiene (i.e. fields in structs are still counted as uncovered), but it's a start.

Would the `--excl-br-line "#\[derive\("` be necessary if the line is already ignored with `--excl-line "#\[derive\("` ?

> Having two parameters is a bit redundant, but I assumed there is a good reason why lcov differentiates between ignored lines and branches.

I'm assuming that we can use the two to discriminate between lines to be completely ignored and lines which need to be counted for coverage but ignore that branches within the line aren't taken.

> `lcovrc` does default to what you indicated, but that means you have ignore patterns whether you want them or not. This smells like an anti-feature to me, but I'm happy to concede to the principle of least surprise and go with what lcov does.

I don't have enough experience here to comment with any authority. Maybe others with some experience could chime in?

I do like drop-in functionality that re-uses prior art and experience. If the thought is that the `LCOV_...` signals would be used by default, a `--no-excl-defaults` option could be added to remove the functionality when desired by the user.

jcdickinson:
Sorry for the late response on this. It has been a crazy week for me.

> So, from some google research and your feedback, it looks like I should just use something like --excl-br-line "assert(_eq|_ne)?!" when calling grcov.

That is correct.

> Would the --excl-br-line "#\[derive\(" be necessary if the line is already ignored with --excl-line "#\[derive\(" ?

Yes, it looks like lcov does this.

> If the thought is that the LCOV_... signals would be used by default, a --no-excl-defaults option could be added to remove the functionality when desired by the user.

Happy to do this, if requested.

.travis.yml
```diff
@@ -126,7 +126,7 @@ script:
...
+        ./target/debug/grcov ccov.zip -s . -t lcov --llvm --branch --ignore-not-existing --ignore "/*" -o lcov.info --excl_line "#\[derive\(" --excl_br_line "#\[derive\(";
```
calixteman, line 129: bad options here, you must replace `_` by `-`

appveyor.yml
```diff
@@ -56,7 +56,7 @@ test_script:
       If ($env:CHANNEL -eq "nightly" -And $env:APPVEYOR_REPO_TAG -eq "false") {
          mkdir ccov_dir
          Get-ChildItem -Path *\grcov*.gc* -Recurse | Copy-Item -Destination ccov_dir
-         .\target\debug\grcov ccov_dir -s . -t lcov --llvm --branch --ignore-not-existing --ignore "C:*" -o lcov.info
+         .\target\debug\grcov ccov_dir -s . -t lcov --llvm --branch --ignore-not-existing --ignore "C:*" -o lcov.info --excl_line "#\[derive\(" --excl_br_line "#\[derive\("
```
calixteman, line 59: same error here

src/file_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+            .enumerate()
+            .into_iter()
+            .filter_map(move |(number, line)| {
+                // Line numbers are 1-based.
+                let number = (number + 1) as u32;
+
+                // The file is split on \n, which may result in a trailing \r
+                // on Windows. Remove it.
+                let line = if line.ends_with("\r") {
```
calixteman, line 75: nit: '\r' instead of the string

src/file_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+        let file = if let Ok(file) = file {
+            file
+        } else {
+            return Vec::new();
+        };
+
+        let mut ignore_br = false;
+        let mut ignore = false;
+
+        file.split("\n")
```
calixteman, line 66: nit: '\n' instead of a string

src/file_filter.rs
```diff
@@ -0,0 +1,129 @@
...
+            .enumerate()
+            .into_iter()
```
calixteman, line 68: nit: probably useless

src/main.rs
```diff
@@ -244,6 +280,33 @@ fn main() {
         }
     };
 
+    let excl_line = matches.value_of("excl-line").map_or(None, |f| {
```
calixteman, line 283: nit: `and_then` instead of `map_or`

src/file_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+            excl_stop: None,
+            excl_br_line: None,
+            excl_br_start: None,
+            excl_br_stop: None,
+        }
+    }
+}
+
+fn matches(regex: &Option<Regex>, line: &str) -> bool {
+    if let Some(regex) = regex {
```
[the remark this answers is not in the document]
calixteman: I'd really prefer `map_or` here

src/file_filter.rs
```diff
@@ -0,0 +1,128 @@
...
+    excl_line: Option<Regex>,
+    excl_start: Option<Regex>,
+    excl_stop: Option<Regex>,
+    excl_br_line: Option<Regex>,
+    excl_br_start: Option<Regex>,
+    excl_br_stop: Option<Regex>,
+}
+
+fn matches(regex: &Option<Regex>, line: &str) -> bool {
+    if let Some(regex) = regex {
```
calixteman, line 21: Why don't you want to use `map_or` ?

jcdickinson: Thanks! I'm still learning Rust and wasn't aware of all the `Option` monads.

calixteman:
@jcdickinson could you update the README to explain how to use it ?

TimDiekmann:
I have used `--excl-br-line "(#\\[derive\\()|(debug_assert)"` which also filters debug assertions, which should never fail, even in production code. This might be added to the README as well.

marco-c:
@TimDiekmann++

TimDiekmann:
Actually I reverted it to `"#\\[derive\\("` and added `-Cdebug-assertions=no` to `RUSTFLAGS` instead.

PR #436, opened 2020-05-08
Improve error message when trying to create the output file

### LINKED ISSUE #435: Output a readable error message when the selected output file is not a file
when i run the example from README with the lcov flag (which is the default), the following happens:
```
» grcov ./target/debug/ -s . --llvm --branch --ignore-not-existing -o ./target/debug/coverage/
10:14:36 [ERROR] A panic occurred at /home/flrn/.cargo/registry/src/github.com-1ecc6299db9ec823/grcov-0.5.13/src/output.rs:24: called `Result::unwrap()` on an `Err` value: Os { code: 21, kind: Other, message: "Is a directory" }
```
i get the same output when i specify `-t lcov` explicitly. it runs fine with the `-t html` flag

calixteman:
Fixes #435.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -21,7 +21,18 @@ use crate::html;
 
 fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
     let write_target: Box<dyn Write> = match output_file {
-        Some(filename) => Box::new(File::create(filename).unwrap()),
+        Some(filename) => {
+            let output = PathBuf::from(filename);
+            if output.is_dir() {
+                panic!(
+                    "The output file {} is a directory and must a regular file.",
+                    filename
+                )
+            }
+            Box::new(File::create(output).unwrap_or_else(|_| {
+                panic!("Cannot create the file {} to dump coverage data", filename)
+            }))
+        }
         None => {
             let stdout = io::stdout();
             Box::new(stdout)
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -548,23 +559,21 @@ mod tests {
...
+                    // 3 hit branches over 10
+                    map.insert(1, vec![true, false, false, true, false, false]);
+                    map.insert(2, vec![false, false, false, true]);
+                    map
                 },
-            ),
-        ];
+                functions: FxHashMap::default(),
+            },
+        )];
```
marco-c, line 576: I guess this is autoformatting? Can you push a commit to reformat everything directly to master, and then rebase this commit on top of it?

src/output.rs
```diff
@@ -21,7 +21,18 @@ use crate::html;
 
 fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
     let write_target: Box<dyn Write> = match output_file {
-        Some(filename) => Box::new(File::create(filename).unwrap()),
+        Some(filename) => {
+            let output = PathBuf::from(filename);
+            if output.is_dir() {
+                panic!(
+                    "The output file {} is a directory and must a regular file.",
```
marco-c, line 28:
```suggestion
                    "The output file {} is a directory, but must be a regular file.",
```

src/output.rs
```diff
@@ -21,7 +21,18 @@ use crate::html;
...
+                    filename
+                )
+            }
+            Box::new(File::create(output).unwrap_or_else(|_| {
+                panic!("Cannot create the file {} to dump coverage data", filename)
+            }))
+        }
```
marco-c, line 35: It'd be nice to do this directly in main, so we don't waste time parsing everything and then error out at the end, but this is OK too.

PR #439, opened 2020-05-11
Use lto for release

calixteman:


marco-c:
Did you try some benchmark to measure the effects?

marco-c:
> Did you try some benchmark to measure the effects?

@calixteman replied on Matrix, the improvement is ~5%.

PR #442, opened 2020-05-13
Add back output-file and mark it as deprecated

calixteman:


marco-c:
Oh I hadn't noticed you published this PR!

PR #446, opened 2020-05-20
Enable deflate feature which is a default backend in zip crate

piotrmaks:
Use `flate2/rust_backend` instead of `flate2/zlib` so there is no additional dependency on external library. It may come handy to keep external library dependencies limited if someone would like to build docker image from scratch just with grcov.

I think that changing default feature in zip is ok but it's also possible to create feature flag and let user decide which backend to use when building.
```
[features]
default = ["deflate-zlib"]
..
deflate-zlib = ["zip/deflate-zlib"]
deflate = ["zip/deflate"]
```

What is your opinion?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,11 @@ exclude = [
 edition = "2018"
 
 [features]
-default = []
+default = ["deflate-zlib"]
 tc = ["tcmalloc"]
+deflate = ["zip/deflate"]
+deflate-miniz = ["zip/deflate-miniz"]
+deflate-zlib = ["zip/deflate-zlib"]
 
 [badges]
 appveyor = { repository = "marco-c/grcov" }
@@ -39,7 +42,7 @@ walkdir = "^2.3"
 num_cpus = "^1.13"
 semver = "^0.10"
 md-5 = "^0.8"
-zip = { version = "^0.4", features = ["deflate"], default-features = false }
+zip = { version = "^0.4", default-features = false }
 tempfile = "^3"
 uuid = { version = "^0.8", features = ["v4"] }
 globset = "^0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ walkdir = "^2.3"
 num_cpus = "^1.13"
 semver = "^0.9"
 md-5 = "^0.8"
-zip = { version = "^0.4", features = ["deflate-zlib"], default-features = false }
+zip = { version = "^0.4", features = ["deflate"], default-features = false }
 tempfile = "^3"
 uuid = { version = "^0.8", features = ["v4"] }
 globset = "^0.4"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -328,6 +328,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "crc32fast 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libz-sys 1.0.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "miniz-sys 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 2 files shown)


marco-c:
Can you check if there is any performance difference? If the performance is the same/better, then I think there is no need to let the user choose.

piotrmaks:
I have modified the benchmark in zip-rs to use the deflate compression method and choose specific backend using a feature flag. Running it on my machine for default Rust backend:
```
cargo bench --no-default-features --features deflate
test read_entry ... bench:     796,345 ns/iter (+/- 54,366) = 1316 MB/s
```
for zlib:
```
cargo bench --no-default-features --features deflate-zlib
test read_entry ... bench:     209,182 ns/iter (+/- 40,825) = 5012 MB/s
```
and for miniz:
```
cargo bench --no-default-features --features deflate-miniz
test read_entry ... bench:     855,404 ns/iter (+/- 63,738) = 1225 MB/s
```

It would be great if someone else could take a look at the benchmark as I am not confident in it.

marco-c:
> Probably running grcov for some huge project would be more realistic.

@calixteman do you have a big zip of Firefox coverage data lying around?

calixteman:
A 43Mb artifact from mozilla CI:
https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/az76xaAIQNCE_y-OpuNI5w/runs/0/artifacts/public/test_info/code-coverage-grcov.zip
you can put it several times in your command line.

calixteman:
I just tryed basic stupid thing:
`time cargo run --release /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip /tmp/code-coverage-grcov.zip -o /dev/null --threads 4` with the above zip
And I consistently get ~14.8s with zllib and ~16.2s without.
So I think we should keep zlib in mozilla ci but I'm fine in general to remove this dependency.
So we could set zlib use in a feature (we already have one for the memory allocator) and by default not use it (like mem allocator).
@marco-c wdyt ?

piotrmaks:
Thanks @calixteman, I've just tried it on my PC (Manjaro 5.6.16-1-MANJARO x86_64, i5-8365U CPU @ 1.60GHz) and for zlib I got ~25s and ~27s for rust backend.
I can update PR to use feature flag for selecting flate2 backend.

marco-c:
WFM!

piotrmaks:
Done.

piotrmaks:
I believe that build on Travis and AppVeyor use the `zlib` backend right now. I set it as a default feature. 

Compiled it on my x86 Linux machine and it looks like it depends on the `libz`
```
ldd target/debug/grcov
        linux-vdso.so.1 (0x00007ffc4c582000)
        libz.so.1 => /usr/lib/libz.so.1 (0x00007f8b0a18d000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f8b0a187000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f8b0a165000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f8b0a14b000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f8b09f84000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f8b0aca7000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f8b09e3f000)
```

Cargo.toml
```diff
@@ -24,8 +24,11 @@ exclude = [
 edition = "2018"
 
 [features]
-default = []
+default = ["deflate-zlib"]
```
marco-c, line 27: I didn't notice you added it as default ;)

Issue #127, opened 2018-05-28
Add clippy in CI


calixteman:

gabriel-v:
Here's what today's clippy outputs with default settings:

```
warning: returning the result of a let binding from a block. Consider returning the expression directly.
  --> src/producer.rs:43:5
warning: this function has too many arguments (9/7)
   --> src/producer.rs:63:5
warning: redundant field names in struct initialization
  --> src/producer.rs:88:29
warning: Err(_) will match all errors, maybe not a good idea
   --> src/producer.rs:201:17
warning: redundant field names in struct initialization
   --> src/producer.rs:244:13
warning: redundant field names in struct initialization
   --> src/producer.rs:245:13
warning: redundant field names in struct initialization
   --> src/producer.rs:284:41
warning: parameter of type `HashMap` should be generalized over different hashers
   --> src/producer.rs:361:40
warning: redundant field names in struct initialization
   --> src/parser.rs:366:29
warning: redundant field names in struct initialization
   --> src/parser.rs:461:25
warning: redundant field names in struct initialization
   --> src/parser.rs:462:25
warning: very complex type used. Consider factoring parts into `type` definitions
   --> src/parser.rs:517:6
warning: this function has too many arguments (9/7)
   --> src/path_rewriting.rs:39:1
warning: the function has a cyclomatic complexity of 51
   --> src/main.rs:44:1
```

This issue depends on these cleanup tasks:

- https://github.com/mozilla/grcov/issues/23 (to reduce main's complexity)
- https://github.com/mozilla/grcov/issues/76 (then add type definitions for those hashmaps)

marco-c:
See also #306.

PR #447, opened 2020-05-20
Pre commit ci

### LINKED ISSUE #302: Use rustfmt and setup CI and pre-commit to use it
The pre-commit configuration can be similar to https://github.com/mozilla/rust-code-analysis/blob/4a899635cc7280bdc265cc2429bd164cfe61714c/.pre-commit-config.yaml.

DanielChabrowski:
Closes #302 

Technically also related to #127, though we can also add a separate nightly clippy job that is allowed to fail and includes the benchmarks. To close it you probably want to fix clippy warnings.

I had to disable benchmarks in clippy because of the:
```error[E0554]: #![feature] may not be used on the stable release channel```

Let me know what you think, cheers.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,7 @@ repos:
         language: system
         types: [file, rust]
         entry: cargo fmt -- --check
+        pass_filenames: false
 
       - id: clippy
         name: clippy
@@ -33,9 +34,9 @@ repos:
       - id: test
         name: test
         language: system
-        files: '\.rs$'
+        types: [file, rust]
         entry: cargo test --lib
         pass_filenames: false
 
 default_language_version:
-    python: python3.7
+    python: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
         name: clippy
         language: system
         types: [file, rust]
-        entry: cargo clippy --all-targets --all # -- -D warnings # Use -D warnings option to ensure the job fails when encountering warnings
+        entry: cargo clippy --bins --tests --examples --all # -- -D warnings # Use -D warnings option to ensure the job fails when encountering warnings
         pass_filenames: false
 
       - id: test
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,7 +123,7 @@ matrix:
             - python3-wheel
       script:
         - rustup component add rustfmt clippy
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: .pre-commit-config.yaml [hunk 1]
```diff
@@ -22,20 +22,21 @@ repos:
         language: system
         types: [file, rust]
         entry: cargo fmt -- --check
+        pass_filenames: false
```
marco-c, line 25: Why is this needed?

.travis.yml
```diff
@@ -109,6 +109,23 @@ matrix:
...
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - python3-pip
+            - python3-setuptools
+            - python3-wheel
+      script:
+        - rustup component add rustfmt clippy
+        - pip3 install pre-commit
```
marco-c, line 126: Nit: use `--disable-pip-version-check --quiet --no-cache-dir` parameters to pip

marco-c:
> I had to disable benchmarks in clippy because of the:
```error[E0554]: #![feature] may not be used on the stable release channel```

We could use Nightly clippy

.pre-commit-config.yaml
DanielChabrowski: Cargo fmt doesn't need passing filenames and `find . -name "*.rs"` includes two somehow unformatted, but also empty, main.rs files in tests.

DanielChabrowski:
> > I had to disable benchmarks in clippy because of the:
> > `error[E0554]: #![feature] may not be used on the stable release channel`
> 
> We could use Nightly clippy exclusively, instead of a Nightly allowed to fail and a stable not allowed to fail.

That's what I actually started with but travis couldn't download nightly fmt. So, since the very beginning it was causing issues. Also, something that "might not work" does not integrate well with pre-commit, it would just be annoying instead of being helpful.

.travis.yml
DanielChabrowski: Done.

.pre-commit-config.yaml
marco-c: Got it!

marco-c:
> That's what I actually started with but travis couldn't download nightly fmt. So, since the very beginning it was causing issues. Also, something that "might not work" does not integrate well with pre-commit, it would just be annoying instead of being helpful.

Fair enough.

PR #455, opened 2020-06-14
Add --keep-only flag to keep specific files/dirs

### LINKED ISSUE #332: Add an option to only keep specific files (opposite of --ignore)
Hi, i use option --ignore <path> to remove paths that i don't want, but linux lcov version has an option to do the oposite:
```
lcov --extract <file> <path*> -o <new file>
```
Create a new lcov info file including information for only that path.

Is there a way to do the same with grcov?

Thanks for the attention and great work.

rdimartino:
Closes #332

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -106,8 +106,7 @@ fn main() {
                                .value_name("PATH")
                                .multiple(true)
                                .number_of_values(1)
-                               .takes_value(true)
-                               .conflicts_with("ignore_dir"))
+                               .takes_value(true))
 
                           .arg(Arg::with_name("path_mapping")
                                .long("path-mapping")
--- a/src/main.rs
+++ b/src/main.rs
@@ -100,6 +100,15 @@ fn main() {
                                .number_of_values(1)
                                .takes_value(true))
 
+                          .arg(Arg::with_name("keep_dir")
+                               .help("Keep only files/directories specified as globs")
+                               .long("keep-only")
+                               .value_name("PATH")
+                               .multiple(true)
+                               .number_of_values(1)
+                               .takes_value(true)
+                               .conflicts_with("ignore_dir"))
+
                           .arg(Arg::with_name("path_mapping")
                                .long("path-mapping")
                                .value_name("PATH")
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -100,6 +100,15 @@ fn main() {
                                .number_of_values(1)
                                .takes_value(true))
 
+                          .arg(Arg::with_name("keep_dir")
+                               .help("Keep only files/directories specified as globs")
+                               .long("keep-only")
+                               .value_name("PATH")
+                               .multiple(true)
+                               .number_of_values(1)
```
calixteman, line 108: I don't think you want that (I fixed it for ignore: https://github.com/mozilla/grcov/pull/459)

src/main.rs
```diff
@@ -100,6 +100,15 @@ fn main() {
...
+                               .takes_value(true)
+                               .conflicts_with("ignore_dir"))
```
calixteman, line 110: I suppose we could have the both options.
For example: keep=foo/*.rs and ignore=foo/bar_*.rs
So having both options could mean: Keep = Keep - (Keep ∩ Ignore)

src/path_rewriting.rs
```diff
@@ -235,6 +236,13 @@ pub fn rewrite_paths(
     }
     let to_ignore_globset = glob_builder.build().unwrap();
 
+    glob_builder = GlobSetBuilder::new();
```
calixteman, line 239: it's the same code to build globs for ignore so you should add a function to do that stuff.

src/main.rs
rdimartino: I chose to make it conflict because that matches the behavior of lcov (`--remove` and `--extract` are conflicting options) and it seemed the least surprising, but I'll take a look at what it would take to implement the set difference as you've described. Thanks!

src/main.rs
calixteman: Please forget my comment, it's probably better like this.

src/main.rs
rdimartino: Ah, I see! This behavior already works as implemented. I'll remove the `conflicts_with` and add a test for using both keep and ignore

calixteman (approved):
Thanks a lot for this patch

PR #459, opened 2020-06-15
Fix option --ignore

calixteman:


calixteman:
Probably better to let like this.

PR #489, opened 2020-10-04
Fix README

MusiKid:
This commit fixes the table of content and introduce some reordering for the titles.


README.md
```diff
@@ -259,7 +275,7 @@ cargo test
 
 ## Minimum requirements
 
-- GCC 4.9 or higher is required (if parsing coverage artifacts generated by GCC).
+* GCC 4.9 or higher is required (if parsing coverage artifacts generated by GCC).
```
marco-c, line 278: Why this change?

marco-c:
Thanks!

README.md
MusiKid: It was just to use the same character everywhere, this isn't really necessary.

marco-c: OK, WFM

marco-c (approved):

PR #512, opened 2020-11-23
Support parsing source-based coverage artifacts

marco-c:


src/llvm_tools.rs
```diff
@@ -0,0 +1,115 @@
+use cargo_binutils::Tool;
```
calixteman, line 1: I don't really see the value of `Tool` here, the code here seems to be pretty simple:
https://docs.rs/cargo-binutils/0.3.3/src/cargo_binutils/tool.rs.html#24
And for example, on my system, I've:
```sh
llvm-profdata-6.0  llvm-profdata-7    llvm-profdata-8    llvm-profdata-9
```
so no `llvm-profdata`.
If possible, it'd be nice to guess the version of the `llvm-profdata` to use in using some bytes in the `.profdata` file.
And then try to call `llvm-profdata` or `llvm-profdata-X` and if something is wrong then print an error message with this information (to avoid bugs about version mismatches between the machine running `grcov` and the one used to produced the `.profdata` files).

calixteman (approved):
If it isn't possible to fix the `llvm-profdata` thing then I'm fine with this patch.

src/llvm_tools.rs
marco-c: cargo_binutils is pretty simple, but I preferred relying on it and not maintain our version of the same code (particularly for https://github.com/rust-embedded/cargo-binutils/blob/115e26e7640337450b609d0d1d14619a1c370c7a/src/rustc.rs, which I feel might change every once in a while).

I like the idea of checking the version, I will file a follow-up issue for that.

marco-c: I filed https://github.com/mozilla/grcov/issues/513.

PR #522, opened 2020-11-26
Update reader.rs to be able to read *all* the recent gcno/gcda files

calixteman:
* it aims to reflect all the recent changes in GCOV.cpp (see llvm tree)
* endianness is correctly detected and handled
* compute counters using spanning tree and Kirchoff's law
* adapt the reader to the version


src/producer.rs
```diff
@@ -713,6 +713,8 @@ mod tests {
                 "rust/generics_with_two_parameters",
                 true,
             ),
+            (ItemFormat::GCNO, true, "reader_gcc-6_1.gcno", true),
+            (ItemFormat::GCNO, true, "reader_gcc-10_1.gcno", true),
```
marco-c, line 717: Could you add gcnos from 7, 8 and 9 too?

marco-c (dismissed):

marco-c (approved):

marco-c:
This caused #672.

PR #523, opened 2020-11-27
Factor out the Command execution

lu-zero:
And make it provide more information in case of error.


marco-c (approved):

marco-c:
LGTM, but the tests need to be updated

lu-zero:
Which ones?

lu-zero:
Updated.

marco-c (approved):

PR #524, opened 2020-11-27
Use all compiled binaries for source-based coverage

marco-c:
This solution isn't perfect as it will run it on all binaries, no matter what, while the best option would be to match the profraw with the right binary.
That's not easy though. We have the possibility of having a "binary signature" in profraw files, but it is not easy to compute: https://github.com/llvm/llvm-project/blob/64ada7accbcccf1ef3ac4910d596e890570c64b2/compiler-rt/lib/profile/InstrProfilingMerge.c#L22-L37 (many of those __llvm_profile functions are platform-specific).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,7 +168,7 @@ pub fn consumer(
     receiver: JobReceiver,
     branch_enabled: bool,
     guess_directory: bool,
-    binary_path: &Option<String>,
+    binary_path: &Option<PathBuf>,
 ) {
     let mut gcov_type = GcovType::Unknown;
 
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -267,7 +267,18 @@ pub fn consumer(
                         binary_path.as_ref().unwrap(),
                         working_dir,
                     ) {
-                        Ok(lcov) => try_parse!(parse_lcov(lcov, branch_enabled), work_item.name),
+                        Ok(lcovs) => {
+                            let mut new_results: Vec<(String, CovResult)> = Vec::new();
+
+                            for lcov in lcovs {
```
calixteman, line 273: nit: map+flatten instead of the loop

calixteman (approved):
lgtm

PR #526, opened 2020-11-27
Parse gcov 9.1 and 10 output and add tests for gcc 10

calixteman:


src/gcov.rs
```diff
@@ -64,50 +64,92 @@ pub fn run_gcov(
...
+            parse_version(&output)
+        };
+    }
+    &V
+}
+
+pub fn get_gcov_output_ext() -> &'static str {
+    lazy_static! {
+        static ref E: &'static str = {
+            if get_gcov_version().major >= 10 {
```
marco-c, line 85: Wasn't it introduced in gcc 9?

src/gcov.rs
```diff
@@ -64,50 +64,92 @@ pub fn run_gcov(
...
+    let mut versions: Vec<_> = gcov_output
+        .split(|c| c == ' ' || c == '\n')
+        .filter_map(|value| Version::parse(value).ok())
+        .collect();
+    assert!(versions.len() >= 1, "no version found for `gcov`.");
+
+    versions.pop().unwrap()
+}
+
+pub fn check_gcov_version() -> bool {
```
marco-c, line 105: Where is this one used other than tests? I couldn't find it

src/parser.rs
```diff
@@ -313,6 +315,51 @@ pub fn parse_lcov(
     Ok(results)
 }
 
+#[derive(Debug, Default, Deserialize)]
+struct GcovJson {
+    format_version: String,
+    gcc_version: String,
+    current_working_directory: Option<String>,
+    data_file: String,
```
marco-c, line 323: What are `current_working_directory` and `data_file` used for?

src/parser.rs
```diff
@@ -324,6 +371,48 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String, CovResult)>, ParserEr
         .unwrap_or_else(|_| panic!("Failed to open gcov file {}", gcov_path.display()));
 
     let mut file = BufReader::new(&f);
+    if gcov_path.extension().unwrap() == "gz" {
```
marco-c, line 374: Let's split the function in two, one for `.json.gz` and one for `gcov`, then check the extension and call the right function, and return an error if the extension is unexpected

src/parser.rs
```diff
@@ -324,6 +371,48 @@ pub fn parse_gcov(gcov_path: &Path) -> Result<Vec<(String, CovResult)>, ParserEr
...
+        let gz = GzDecoder::new(file);
+        let mut gcov: GcovJson = serde_json::from_reader(gz).unwrap();
```
marco-c, line 376: Maybe we should print a warning if the `format_version` is not the one we expect

src/gcov.rs
calixteman: I've no idea, I let this code because it was there...

src/parser.rs
calixteman: `current_working_directory` is the output of `getpwd()` when the gcno is generated:
https://github.com/gcc-mirror/gcc/blob/e87559d202d90e614315203f38f9aa2f5881d36e/gcc/coverage.c#L1321
`data_file` is the gave as argument of `gcov`

src/gcov.rs
marco-c: Let's file an issue to remove it then

src/gcov.rs
calixteman: yep correct

src/lib.rs
```diff
@@ -198,7 +202,14 @@ pub fn consumer(
                         }
 
                         let mut new_results = if gcov_type == GcovType::SingleFile {
-                            let new_results = try_parse!(parse_gcov(&gcov_path), work_item.name);
+                            let new_results = try_parse!(
+                                if gcov_ext.ends_with("gz") {
+                                    parse_gcov_gz(&gcov_path)
+                                } else {
```
marco-c, line 208: else endswith("gcov")
parse_gcov
else
log error

src/producer.rs
```diff
@@ -118,7 +118,8 @@ impl Archive {
     fn is_gcno_llvm(reader: &mut dyn Read) -> bool {
         let mut bytes: [u8; 8] = [0; 8];
         reader.read_exact(&mut bytes).is_ok()
-            && bytes == [b'o', b'n', b'c', b'g', b'*', b'2', b'0', b'4']
+            && &bytes[..5] == b"oncg*"
+            && (&bytes[5..] == b"204" || &bytes[5..] == b"804")
```
marco-c, line 122: Why this change?

src/parser.rs
marco-c: Can you add this as a comment to explain the fields?

src/producer.rs
calixteman: to have integration test green with clang 11

src/parser.rs
```diff
@@ -313,6 +315,105 @@ pub fn parse_lcov(
...
+    let f = File::open(&gcov_path)
+        .unwrap_or_else(|_| panic!("Failed to open gcov file {}", gcov_path.display()));
+
+    let file = BufReader::new(&f);
+    let gz = GzDecoder::new(file);
+    let mut gcov: GcovJson = serde_json::from_reader(gz).unwrap();
+    let mut results = Vec::new();
+
+    if gcov.format_version != "1" {
+        eprintln!(
```
marco-c, line 373: We usually use `log::error` for printing errors

marco-c (approved):

src/lib.rs
calixteman: the test is made on `gcov_ext` set to either "gcov.json.gz" or "gcov" so there is no other alternative.

marco-c: That's true now, but who knows how the code will evolve, the check would allow us to be more defensive.
It's a nit though, feel free to ignore ;)

marco-c (approved):
LGTM!

Don't forget about filing the issue for removing check_gcov_version

marco-c (approved):

marco-c (approved):

PR #527, opened 2020-11-28
Add an option to demangle function names

calixteman:


src/main.rs
```diff
@@ -473,9 +478,9 @@ fn main() {
     );
```
marco-c, line 479: Instead of applying the demangling in each output function, it'd be preferable to apply it to the result first, then pass the demangled result to the output functions

marco-c: Otherwise we have to duplicate the demangling support in all output functions

marco-c:
I guess as part of this you can remove this TODO: https://github.com/mozilla/grcov/blob/5de5d4dedb0ae0b15b87b0849fcdec9e0280c124/src/llvm_tools.rs#L57.

src/main.rs
```diff
@@ -235,7 +235,11 @@ fn main() {
                                 .help("Marks the end of a section excluded from branch coverage. The current line is part of this section.")
                                 .long("excl-br-stop")
                                 .value_name("regex")
-                                .takes_value(true))
+                                 .takes_value(true))
+
+                            .arg(Arg::with_name("demangle")
+                               .help("Demangle cpp and Rust symbols name")
+                               .long("demangle"))
```
marco-c, line 242: What if we made this true by default?

src/main.rs
calixteman: Some outputs don't need to have demangled symbols... so I added this feature only in outputs which outputs function name to avoid useless stuff for other outputs.

src/main.rs
calixteman: I'm fine with a true by default, so let's rename this option into `no-demangle` or something else.

src/main.rs
```diff
@@ -235,7 +235,11 @@ fn main() {
                                 .help("Marks the end of a section excluded from branch coverage. The current line is part of this section.")
                                 .long("excl-br-stop")
                                 .value_name("regex")
-                                .takes_value(true))
+                                 .takes_value(true))
```
marco-c, line 238: Nit: you added an extraneous space here (before the ".")

marco-c (dismissed):

marco-c (approved):

Issue #525, opened 2020-11-27
Handle overflows in counters


marco-c:
Revert 155c73f05c67f173c75946b07e22b9d53f65d647, and correctly handle overflows in counters (apparently there are lines which are hit many many times).

marco-c:
STR:
```
export RUSTFLAGS="-Zinstrument-coverage"
LLVM_PROFILE_FILE="grcov-%p-%m.profraw" cargo test
cargo run -- `find . \( -name "grcov-*.profraw" \) -print` --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" -o lcov.info
```

marco-c:
This shows something is wrong with counters:
```
export RUSTFLAGS="-Zinstrument-coverage"
LLVM_PROFILE_FILE="grcov-%p-%m.profraw" cargo test --lib path_rewriting
rust-profdata merge -sparse `find . \( -name "grcov-*.profraw" \) -print` -o grcov.profdata
rust-cov export target/debug/deps/grcov-13778df2ccd171b3 --instr-profile grcov.profdata --format lcov
```
N.B.: `grcov-13778df2ccd171b3` should be the binary that is run as part of cargo test.

You will see some lines which have unexpectedly high counts, such as `DA:119,18446744073709551615`.

calixteman:
The number `18446744073709551615` is `(1 << 64 - 1)` so it looks like it's a `int64(-1)` casted into a uint64.

PR #530, opened 2020-11-30
Use saturating_add in merge_results

lu-zero:


calixteman:
I don't think it's a good idea.
Counters are u64, so in normal conditions it's impossible to reach the max... and so it's very likely a good way to hide a real problems and maybe show inaccurate results.

marco-c:
@calixteman between the two options (overflowing and risking a 0 counter and saturating), the saturating option seems better to me.
We can print a warning in case the add would overflow to avoid "hiding" the problem.

marco-c:
@lu-zero as part of this, could you also revert 155c73f05c67f173c75946b07e22b9d53f65d647 (thus fixing #525).

calixteman:
I'm fine with having an error at least.
I wonder if -1 is more or less equivalent `to counter is 0` or `no counter`.
I'd say have a look on where are these bad counters and then choose on of the above options.
And if none of the above options is correct then use `saturating_add` (with a TODO) is fine.
But we really need to investigate either on llvm's side or rustc's one where is the problem.

lu-zero:
To reproduce the problem you can run gcovr on the gcovr tests :) Let me know how you'd like to proceed.

marco-c:
@lu-zero let's:
1. print a warning if there could be an overflow;
2. don't overflow but saturate.

Also get rid of the workaround from 155c73f05c67f173c75946b07e22b9d53f65d647 which should no longer be necessary.

src/lib.rs
```diff
@@ -65,7 +65,11 @@ pub fn merge_results(result: &mut CovResult, result2: CovResult) {
     for (&line_no, &execution_count) in &result2.lines {
         match result.lines.entry(line_no) {
             btree_map::Entry::Occupied(c) => {
-                *c.into_mut() += execution_count;
+                if c.get().checked_add(execution_count).is_none() {
+                    warn!("Execution count overflow detected.");
```
calixteman, line 69: warn only one time.

src/lib.rs
```diff
@@ -65,7 +65,11 @@ pub fn merge_results(result: &mut CovResult, result2: CovResult) {
...
+                }
+
+                *c.into_mut() = c.get().saturating_add(execution_count);
```
calixteman, line 72: if add is correct from the above check then no need to call saturating_add.

Cargo.toml
```diff
@@ -66,7 +66,7 @@ flate2 = "^1.0"
 lazy_static = "^1.3"
 
 [profile.release]
-lto = true
+lto = "thin"
```
calixteman, line 69: why this change ?

lu-zero: Marco asked for a patch about it, it ended up in this set, I can make a separate PR for it.

src/lib.rs
```diff
@@ -61,11 +61,17 @@ use std::path::PathBuf;
...
-pub fn merge_results(result: &mut CovResult, result2: CovResult) {
+pub fn merge_results(result: &mut CovResult, result2: CovResult) -> bool {
+    let mut warn_overflow = false;
     for (&line_no, &execution_count) in &result2.lines {
         match result.lines.entry(line_no) {
             btree_map::Entry::Occupied(c) => {
-                *c.into_mut() += execution_count;
+                let v = c.get().checked_add(execution_count).unwrap_or_else(|| {
+                    warn_overflow = true;
+                    c.get().saturating_add(execution_count)
```
calixteman, line 71: in this case just return u64::max: no need to call saturating_add

calixteman (approved):

Issue #283, opened 2019-04-27
Use rust hashmap instead of fxhashmap


calixteman:
Recently, hashbrown (https://github.com/rust-lang/hashbrown) has been included in rust-lang so according to the benchmark (https://github.com/rust-lang/hashbrown#performance) we should at some point evaluate and probably use the rust HashMap.

eddyb:
`FxHashMap` is the `std` `HashMap`, with a different hasher (cheaper and deterministic, at the cost of removing DoS protection).

I'm not sure anything needs to be done here, I can't imagine the default hasher being faster.

PR #532, opened 2020-12-01
New Html output

lu-zero:
It is using Tera templates so in the future we can let the user pass custom ones.

image
image
image


gilescope:
Just want to say <3 the improved css - the last .css file was like a blast of the '90s.

src/templates/macros.html
```diff
@@ -0,0 +1,59 @@
+{% macro summary_line(kind, covered, total) %}
+    {% set per = percent(num=covered, den=total) %}
+    <div class="level-item has-text-centered">
+        <div>
+            <p class="heading">{{ kind | capitalize }}</th>
```
calixteman, line 5: You must run a html validator against the different possible generated htmls to be sure that everything is fine.
For example it'd have helped you to get a spurious `</th>` here.

src/templates/file.html
```diff
@@ -0,0 +1,38 @@
+{% import "macros.html" as macros %}
```
calixteman, line 1: I ran that stuff on grcov itself and open for example `parser.rs.html` and there are a lot of useless `\n` or `<space>`.
It'd be nice to strip them to reduce the size of generated files.

src/html.rs
```diff
@@ -261,7 +184,26 @@ fn get_dirs_result(global: Arc<Mutex<HtmlGlobalStats>>, rel_path: &PathBuf, stat
     };
 }
 
-pub fn gen_index(global: HtmlGlobalStats, conf: Config, output: &PathBuf, branch_enabled: bool) {
+use tera::{Context, Tera};
+
+fn make_context() -> Context {
+    let mut ctx = Context::new();
+
+    if let Ok(ver) = std::env::var("BULMA_VERSION") {
```
calixteman, line 192: nit: use `Result::map_or` here

src/html.rs
```diff
@@ -1,25 +1,15 @@
 use chrono::{DateTime, Utc};
-use fomat_macros::fomat;
-use std::cmp::Ordering;
-use std::collections::{btree_map, BTreeSet};
+use serde_json::value::{from_value, to_value, Value};
+use std::collections::HashMap;
```
calixteman, line 3: nit: we use FxHashMap everywhere so do the same here.
Anyway if you see any good reason to use a collections::HashMap here and in other places, please do that in an other patch.

marco-c: There's #283 which we filed a while ago, it'd be nice to finally fix it (switch everything to HashMap and drop FxHashMap).

marco-c: (I agree, in a separate patch)

src/templates/macros.html
```diff
@@ -0,0 +1,59 @@
+{% macro summary_line(kind, covered, total) %}
+    {% set per = percent(num=covered, den=total) %}
```
calixteman, line 2: To be clear I don't know `tera` so maybe my next questions are useless.
Isn't overkill to call `percent` function using a `HashMap` ?
Isn't possible to just call `percent(covered, total)` ?
`percent` is just an example here

calixteman (commented):
Looks good overall.Thanks a lot for doing this, it looks a way nicer now.

src/templates/macros.html
lu-zero: It is not possible, tera uses that way to pass the arguments.

calixteman: Ok. It's kinda stupid... because finally we could just pass the arguments in an array and an array is finally a hashmap with keys equal to 0 & 1 here...

src/templates/macros.html
lu-zero: I'll do.

src/html.rs
lu-zero: tera traits use HashMap.

calixteman: I saw that: https://docs.rs/tera/1.5.0/tera/trait.Function.html

src/html.rs
lu-zero: Sure.

src/templates/macros.html
lu-zero: tidy-html5 seems happy now

lu-zero:
Rebased and with, hopefully, all the issues addressed.

src/templates/file.html
```diff
@@ -0,0 +1,38 @@
...
+{% extends "base.html" %}
```
calixteman, line 2: nit: 
image
It'd be nice to have no warning: see https://validator.w3.org/ to check yourself

calixteman (commented):
There is grcov.css file to remove too which is useless I guess (and still used in html.rs)

src/templates/file.html
lu-zero: It had been addressed in the updated set.

PR #533, opened 2020-12-02
WIP: Initial cargo-grcov command

gilescope:
WIP PR for cargo-grcov

I'm a big fan of early review to make sure I'm headed in the right direction. Wasn't qute sure how grcov created errors - they seem to be logged with log.error!(), but didn't spot any custom errors - should I just create a specific error type where needed - or is there one that converts from a string?

TODO:
   [ ] At the moment it doesn't detect that the tests are already run and runs them again.

### DIFF
--- a/src/bin/cargo-grcov.rs
+++ b/src/bin/cargo-grcov.rs
@@ -1,0 +1,513 @@
+extern crate clap;
+
+use std::collections::HashMap;
+use std::ffi::OsString;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+
+use clap::{crate_authors, crate_version, App, AppSettings, Arg, SubCommand};
+
+type Env = HashMap<OsString, OsString>;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct Context {
+    pwd: PathBuf,
+    args: Vec<OsString>,
+    env: Env,
+}
+
+impl Default for Context {
+    fn default() -> Self {
+        Self {
+            pwd: std::env::current_dir().unwrap(),
+            args: std::env::args_os().collect(),
+            env: std::env::vars_os().collect(),
+        }
+    }
+}
+
+fn main() {
+    let context = Context::default();
+
+    match parse_args(context) {
+        Ok(actions) => {
+            if let Err(err) = acts(&actions) {
+                eprintln!("Error executing: {}", err);
+                std::process::exit(-2);
+            }
+        }
+        Err(err) => {
+            eprintln!("Error parsing: {}", err);
+            std::process::exit(-1);
+        }
+    }
+}
+
+fn acts(actions: &Vec<Action>) -> Result<(), Box<dyn std::error::Error>> {
+    for action in actions {
+        let mut cmd = act(&action);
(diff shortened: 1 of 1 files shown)


marco-c:
Sorry for the delay, I hope I'll be able to get to this by the end of this week.

dnaka91:
Is there any progress on this?

I'm recently using grcov a lot and it would be great to have a dedicated `cargo-grcov` command that does all of the heavy lifting within a single command. Currently have to copy over a bunch of setup scripts for each project and this would simplify things A LOT.

marco-c:
Unfortunately I didn't have time to review this :(

If somebody can help me with reviewing the code in detail, I can review the overall architecture.

(Assuming @gilescope is still interested in finishing this).

dnaka91:
I looked through the code and it is pretty much a wrapper around the `grcov` binary with the arguments and env vars described in the README.md directly applied by this new binary. It seems to miss support for source-based coverage yet.

I personally prefer to have `cargo-grcov` being a second binary that is independent of the `grcov` binary and simply uses the same components directly. That avoids version mismatches and failures trying to resolve the location of `grcov`.

Pretty much what was done for the [flamegraph](https://github.com/flamegraph-rs/flamegraph/blob/master/src/bin/cargo-flamegraph.rs) tool is what I think is more effective and safer towards errors. As we have a good abstraction already, it's easy to implement as well.

If you, @marco-c, don't mind, I would start a new PR that uses the same technique as in the flamegraph crate and combine it with the effort already done here, then adding in more features. For that PR I would probably strip many options from the CLI and only expose what make sense in the context of cargo.

What do you think?

marco-c:
@dnaka91 sounds good to me!

marco-c:
@dnaka91 sounds good to me!

gilescope:
Sounds good to me too - I just want to use it :-) so happy to test out,
feedback etc.

marco-c:
@dnaka91 still interested?

dnaka91:
Definitely! Sorry, I didn't post any update on this for a while. Had a few other things going on and simply didn't get to it yet. Already checked the current implementation of this PR and I think I'll go with a very similar approach and then add a few cli args from the original grcov binary.

Also, I was hoping that #648 would maybe be merged before I start on this (or decided not to, just so I don't have to implement the CLI logic twice). Is there any chance you could give it a look soonish? 🙇‍♂️

gilescope:
@dnaka91 structopt is merged in now if that helps...

gilescope:
I've updated this PR to structopt. @dnaka91 as I said if you want to go on from here that's cool with me.

dnaka91:
@gilescope thanks for letting me know. I didn't work on this for quite a long time but will likely be able to pick up on this again soon-ish.

gilescope:
Great news. Happy to test out and give feedback.

marco-c:
@dnaka91 @gilescope does any of you want to pick this up again?

dnaka91:
Sorry, I didn't do any further work on this. Reason is, that I moved to `cargo-llvm-cov` which just works better for me and seems more active compared to grcov. Therefore, I don't really have the motivation to push this further.

PR #536, opened 2020-12-03
Remove fxhashmap

### LINKED ISSUE #283: Use rust hashmap instead of fxhashmap

lu-zero:
Fixes #283.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,14 +23,9 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "ahash"
-version = "0.6.1"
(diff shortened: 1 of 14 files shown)


marco-c:
@lu-zero could you paste here the results of the benchmarks that you run?

eddyb:
As per https://github.com/mozilla/grcov/issues/283#issuecomment-909239441, I don't think anything needs to be done here.

lu-zero:
Yes, I confirmed on matrix but I forgot to update this PR.

PR #538, opened 2020-12-03
Unbreak appveyor

lu-zero:


marco-c (approved):
This is not fixing all issues with Appveyor, but it's still good to land.

Issue #483, opened 2020-09-02
Subfolder will be missed in reports if it has the same name as the root folder


ko1N:
In my current Rust project I have a workspace that contains a multiple sub-crates.

The git repository is named "memflow" (and so is my folder on disk) and one of the crates inside the project workspace is also named "memflow". In this case grcov will skip over the entire memflow crate for me and not provide any coverage on it whatsoever.

If i rename the root folder on disk to something like memflow_test so that the main crate is at memflow_test/memflow grcov will pick it up.

The project in question can be found here:
https://github.com/memflow/memflow/

I was running grcov as described in the Readme. I additionally added the "--workspace --all-features" flags for the cargo build & cargo test steps.

Best regards,
Patrick

tylerhawkes:
I'm seeing the same thing with a properietary repo.

vitorenesduarte:
I've described the same issue in https://github.com/actions-rs/grcov/issues/73. There I mentioned (https://github.com/actions-rs/grcov/issues/73#issuecomment-635205701) that the issue was not occurring when using Travis. But maybe it was because the format I was using in Travis was `coveralls`, not `lcov`.

Could this issue be specific to the `lcov` format?

vitorenesduarte:
I believe https://github.com/mozilla/grcov/pull/546 fixes this. It seems that the issue was not specific to any format.

PR #546, opened 2020-12-13
Do not ignore subfolder with name equal to root

vitorenesduarte:
Before this commit, a subfolder with the same name as the root/source folder would be ignored from the coverage reports (see https://github.com/mozilla/grcov/issues/483). This was occurring because the ancestor of the subfolder was being pruned (because it matched the end of the source folder) in the method used to guess the absolute path. This resulted in an incorrect guess of the absolute path.

With this commit, we first check whether the join of the subfolder relative path with the source folder exists, and if so, return it as the guess of the absolute path. If the join doesn't exist, we use the previous guess method.

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -91,14 +91,20 @@ fn apply_mapping(mapping: &Option<Value>, path: &str) -> PathBuf {
     PathBuf::from(path)
 }
 
-// Remove common part between the prefix's end and the path's start
+// If the join of the source and the relative path is a file, return it.
+// Otherwise, remove common part between the source's end and the relative
+// path's start.
 fn guess_abs_path(prefix_dir: &PathBuf, path: &PathBuf) -> PathBuf {
+    let full_path = prefix_dir.join(path);
+    if full_path.is_file() {
+        return full_path;
+    }
     for ancestor in path.ancestors() {
         if prefix_dir.ends_with(ancestor) && !ancestor.as_os_str().is_empty() {
             return prefix_dir.join(path.strip_prefix(ancestor).unwrap().to_path_buf());
         }
     }
-    prefix_dir.join(path)
+    full_path
 }
 
 // Remove prefix from the source file's path.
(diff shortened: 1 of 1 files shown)


src/path_rewriting.rs
```diff
@@ -91,14 +91,20 @@ fn apply_mapping(mapping: &Option<Value>, path: &str) -> PathBuf {
...
-// Remove common part between the prefix's end and the path's start
-fn guess_abs_path(prefix_dir: &PathBuf, path: &PathBuf) -> PathBuf {
-    for ancestor in path.ancestors() {
-        if prefix_dir.ends_with(ancestor) && !ancestor.as_os_str().is_empty() {
-            return prefix_dir.join(path.strip_prefix(ancestor).unwrap().to_path_buf());
+// If the join of the source and the relative path is a file, return it.
+// Otherwise, remove common part between the source's end and the relative
+// path's start.
+fn guess_abs_path(source_dir: &PathBuf, rel_path: &PathBuf) -> PathBuf {
+    let join = source_dir.join(rel_path);
```
marco-c, line 98: Nit: ```suggestion
    let full_path = source_dir.join(rel_path);
```

File: src/path_rewriting.rs [hunk 2]
```diff
@@ -91,14 +91,20 @@ fn apply_mapping(mapping: &Option<Value>, path: &str) -> PathBuf {
...
 
-// Remove common part between the prefix's end and the path's start
-fn guess_abs_path(prefix_dir: &PathBuf, path: &PathBuf) -> PathBuf {
-    for ancestor in path.ancestors() {
-        if prefix_dir.ends_with(ancestor) && !ancestor.as_os_str().is_empty() {
-            return prefix_dir.join(path.strip_prefix(ancestor).unwrap().to_path_buf());
+// If the join of the source and the relative path is a file, return it.
+// Otherwise, remove common part between the source's end and the relative
+// path's start.
+fn guess_abs_path(source_dir: &PathBuf, rel_path: &PathBuf) -> PathBuf {
```
marco-c, line 97: Nit: keep the same argument names as before, so the diff is smaller

vitorenesduarte:
@marco-c I've addressed your comments. (I force-pushed as the diff is small.)

src/path_rewriting.rs
```diff
@@ -91,14 +91,20 @@ fn apply_mapping(mapping: &Option<Value>, path: &str) -> PathBuf {
...
+        return full_path;
+    }
     for ancestor in path.ancestors() {
         if prefix_dir.ends_with(ancestor) && !ancestor.as_os_str().is_empty() {
             return prefix_dir.join(path.strip_prefix(ancestor).unwrap().to_path_buf());
         }
     }
-    prefix_dir.join(path)
+    full_path
 }
```
marco-c, line 108: I was thinking of an alternative solution to avoid any I/O (`is_file` does I/O).

Instead of stopping at the first ancestor for which `prefix_dir.ends_with(ancestor)`, what if we stop at the last one?

vitorenesduarte: I don't fully grasp the purpose of stripping here, but I don't think that that would work in the case this PR tries to address. For example,  in the test case I added, the source directory `S`  and the relative path `P` have stuff in common, but what we want is really `S + P`, i.e. no stripping.

vitorenesduarte: To try to understand `guess_abs_path`, I've changed its implementation to simply be `prefix_dir.join(path)` (so, no stripping) and the tests seem to pass. @marco-c  Could it be that we don't need to perform any stripping here?

marco-c: Right, I see no other way.

BTW, the `guess_abs_path` function was introduced in #197. I don't recall exactly what was the real world problem it fixed.

marco-c (approved):

src/path_rewriting.rs
marco-c: > To try to understand guess_abs_path, I've changed its implementation to simply be prefix_dir.join(path) (so, no stripping) and the tests seem to pass. @marco-c Could it be that we don't need to perform any stripping here?

Interesting, I thought the test added as part of #197 would fail without that "guess"!

vitorenesduarte: From what I understand, the added test is not testing the guessing part. It seems that that test passes due to the `assert` in the added 
```rust
    if let Some(p) = &source_dir {
        assert!(p.is_absolute());
    }
```
failing.

PR #548, opened 2020-12-17
Remove absolute path guessing.

vitorenesduarte:
Absolute path guessing was introduced in https://github.com/mozilla/grcov/pull/197 but the discussion in https://github.com/mozilla/grcov/pull/546 suggests that it may not be needed. This commit removes such mechanism.


marco-c:
Could you rebase the PR?

PR #599, opened 2021-04-09
Add cobertura output option

### LINKED ISSUE #468: Generate coverage output for Gitlab/Cobertura?
Would it be possible to add support for generating Cobertura xml coverage reports? [Gitlab's coverage reporting](https://docs.gitlab.com/ee/user/project/merge_requests/test_coverage_visualization.html) uses Cobertura's format.

It appears as though [tarpaulin supports cobertura](https://github.com/xd009642/tarpaulin#pycobertura), so it would be good if `grcov` could too.

geovie:
Hi,
this PR adds a new output option for the cobertura xml format (as defined in http://cobertura.sourceforge.net/xml/coverage-04.dtd).

I only did some basic testing, so I'm not sure how correct the output is, but it should be a good starting point for now.

Fixes #468

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,8 +107,9 @@ OPTIONS:
             - *covdir* for the covdir recursive JSON format;
             - *coveralls+* for the Coveralls specific format with function information;
             - *ade* for the ActiveData-ETL specific format;
+            - *cobertura* for a cobertura coverage report;
             - *files* to only return a list of files.
-             [default: lcov]  [possible values: ade, lcov, coveralls, coveralls+, files, covdir, html]
+             [default: lcov]  [possible values: ade, lcov, coveralls, coveralls+, files, covdir, html, cobertura]
         --path-mapping <PATH>...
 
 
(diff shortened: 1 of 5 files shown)


marco-c (commented):
I didn't do a super careful review since it's a new output format, but it looks good overall.
Can you add some basic tests though? You can write tests for the `output_cobertura` function directly, making sure the cobertura output is what you'd expect given the coverage data you pass in.

geovie:
@marco-c Added some tests and fixed the coverage stats.
Previously it could happen that the stats were wrong if e.g. two function with the same lines are reported the lines would be counted twice.

marco-c (approved):
Thanks!

marco-c:
@geovie would you like to document how to use grcov with GitLab?

svenstaro:
I'd much appreciate a new release of grcov with this PR in it. :)

marco-c:
@svenstaro I've released 0.8.0.

svenstaro:
Super nice, thanks!

PR #604, opened 2021-04-19
Add a clearer error message when output fails

### LINKED ISSUE #481: Add a clear error message when output fails if directory doesn't exist
```
panic <...> grcov-0.5.15\src\output.rs:33: 
Cannot create the file .\coverage\lcov.info to dump coverage data
```
I'll create directory manually for now

mariot:
because target directory doesn't exist

Fix #481

Add a clear error message when output fails if directory doesn't exist

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -45,8 +45,17 @@ pub fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
                     filename
                 )
             }
-            Box::new(File::create(output).unwrap_or_else(|_| {
-                panic!("Cannot create the file {} to dump coverage data. Check if parent directory exists.", filename)
+            Box::new(File::create(&output).unwrap_or_else(|_| {
+                let parent = output.parent();
+                if let Some(filepath) = parent {
+                    if !filepath.exists() {
+                        panic!(
+                            "Cannot create {} to dump coverage data, as {} doesn't exist",
+                            filename, filepath.display()
+                        )
+                    }
+                }
+                panic!("Cannot create the file {} to dump coverage data.", filename)
             }))
         }
         None => {
--- a/src/output.rs
+++ b/src/output.rs
@@ -46,7 +46,7 @@ pub fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
                 )
             }
             Box::new(File::create(output).unwrap_or_else(|_| {
-                panic!("Cannot create the file {} to dump coverage data", filename)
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -46,7 +46,7 @@ pub fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
                 )
             }
             Box::new(File::create(output).unwrap_or_else(|_| {
-                panic!("Cannot create the file {} to dump coverage data", filename)
+                panic!("Cannot create the file {} to dump coverage data. Check if parent directory exists.", filename)
```
marco-c, line 49: This error isn't only printed when the parent directory doesn't exist. Instead of asking the user to check, grcov can check first if the parent directory exists and panic if it doesn't with a clearer error message "Cannot create path/to/file to dump coverage, as path/to/ doesn't exist".

mariot: Ah great. I was thinking about that. I'm changing it!

mariot: The code I wrote looks very dirty. How would you ameliorate it?

I also tried: 
```rust
output.pop();
if !output.exists() {
    panic!(
        "Cannot create {} to dump coverage data, as {} doesn't exist",
        filename, output.display()
    )
}
```
but we need to make output mutable

and
```rust
match output.parent() {
                    Some(parent_path) => {
                        if !parent_path.exists() {
                            panic!(
                                "Cannot create {} to dump coverage data, as {} doesn't exist",
                                filename, parent_path.display()
                            )
                        }
                    },
                    None => {}
                };
```

and 

```rust
let parent = output.parent().unwrap();
                if !parent.exists() {
                    panic!(
                        "Cannot create {} to dump coverage data, as {} doesn't exist",
                        filename, parent.display()
                    )
                }
```

src/output.rs
```diff
@@ -45,8 +45,17 @@ pub fn get_target_output_writable(output_file: Option<&str>) -> Box<dyn Write> {
                     filename
                 )
             }
-            Box::new(File::create(output).unwrap_or_else(|_| {
-                panic!("Cannot create the file {} to dump coverage data", filename)
+            Box::new(File::create(&output).unwrap_or_else(|_| {
+                let parent = output.parent();
+                if let Some(filepath) = parent {
```
marco-c, line 50: Nit: 
```suggestion
                if let Some(parent_path) = parent {
```

marco-c (dismissed):
> The code I wrote looks very dirty

It doesn't look that bad!

mariot:
Cool, I pushed the last commit

marco-c (approved):

PR #620, opened 2021-05-17
Avoid calling `Tool::Cov.path` multiple times

EFanZh:
A single `Tool::Cov.path()` call takes more than 300 ms to finish on my Windows PC on average. I have found out that for one of my projects, the total time took by `Tool::Cov.path()` is more than 30 seconds, which is too slow, so I think the call should stay outside of the loop to perform better.

This PR also fixes some Clippy warnings, replaces some `str` with `OsStr` to avoid unnecessary `unwrap()` calls, and fixes a source code format error.

### DIFF
--- a/src/llvm_tools.rs
+++ b/src/llvm_tools.rs
@@ -1,11 +1,11 @@
 use cargo_binutils::Tool;
 use is_executable::IsExecutable;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use walkdir::WalkDir;
 
-pub fn run(cmd: PathBuf, args: &[&str]) -> Result<Vec<u8>, String> {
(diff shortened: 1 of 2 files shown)


marco-c:
@EFanZh LGTM, but could you split the PR into multiple logically separated commits? E.g. 01d10fd4cd6f1c0803e85b1eeef5533a4962bf38 doesn't only contain what it says on the tin :)

EFanZh:
I have updated my commits, but why is the CI failing?

marco-c (approved):
Thanks!

marco-c:
> I have updated my commits, but why is the CI failing?

It's not your fault, they are pre-existing issues.
We have plans to switch to another CI system, Taskcluster, but never got to it.

PR #622, opened 2021-05-19
Simplify checking executable files

EFanZh:
`is_executable` makes sure that `entry.path()` is a file, so we don’t need to check it manually: https://github.com/fitzgen/is_executable/pull/4.

Also, the `to_path_buf()` call is also unnecessary since we can get the full path from the `entry` object.


marco-c (approved):

marco-c:
Thanks!

PR #625, opened 2021-05-27
Fix all reported clippy warnings

dnaka91:
Currently clippy reports many warnings. This PR fixes all of them with the exception of a  
few being allowed now as it bring any benefit to adjust to them.

While doing so I replaced many usages of `PathBuf` with `&Path` as well.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -96,7 +96,7 @@ pub fn output_activedata_etl(results: CovResultIter, output_file: Option<&str>, 
         for function in result.functions.values() {
             start_indexes.push(function.start);
         }
-        start_indexes.sort();
+        start_indexes.sort_unstable();
 
         for (name, function) in &result.functions {
             // println!("{} {} {}", name, function.executed, function.start);
@@ -227,10 +227,10 @@ pub fn output_covdir(results: CovResultIter, output_file: Option<&str>) {
         ));
     }
 
-    let mut global = global.borrow_mut();
+    let mut global = global.take();
     global.set_stats();
 
-    serde_json::to_writer(&mut writer, &global.to_json()).unwrap();
+    serde_json::to_writer(&mut writer, &global.into_json()).unwrap();
 }
 
 pub fn output_lcov(results: CovResultIter, output_file: Option<&str>, demangle: bool) {
@@ -336,7 +336,7 @@ where
         .and_then(|child| child.wait_with_output())
         .ok() // Discard the error type -- we won't handle it anyway
         .and_then(|output| String::from_utf8(output.stdout).ok())
-        .unwrap_or_else(|| String::new())
+        .unwrap_or_default()
 }
 
 /// Returns a JSON object describing the given commit. Coveralls uses that to display commit info.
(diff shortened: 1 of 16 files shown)


src/reader.rs
```diff
@@ -361,12 +361,11 @@ impl Debug for GCNO {
...
 }
 
-impl GCNO {
+impl Gcno {
     pub fn new() -> Self {
-        GCNO {
+        Gcno {
             version: 0,
             checksum: 0,
-            cwd: None,
```
calixteman, line 369: I'm not sure that's a good idea to remove this property even if it's unused.

dnaka91: I stumbled over this at first and quickly got some test failures.

Everything is good as long as the value for still read [here](https://github.com/mozilla/grcov/pull/625/files#diff-267c9f4da66412a9f439ac08d224356fe24265b5e1cebb6c44c2d55b96414513R454). I simply removed it from the struct but the reading/parsing logic _should_ not be affected.

Of course I can put it back in if you prefer.

calixteman: Yep I saw that the value is still read.
It'd prefer keeping it in case we need it or even for debug purposes.

dnaka91: I put it back in now. Interestingly clippy didn't warn about it anymore now 🤔

src/output.rs
```diff
@@ -96,7 +96,7 @@ pub fn output_activedata_etl(results: CovResultIter, output_file: Option<&str>,
         for function in result.functions.values() {
             start_indexes.push(function.start);
         }
-        start_indexes.sort();
+        start_indexes.sort_unstable();
```
calixteman, line 99: I suppose that all `function.start` are different so I guess that using `sort_unstable` doesn't make any difference in the final order with `sort`. Out of curiosity, do you have a link on some benchmarks to compare perf for an array with unique elements ?
Anyway we should maybe use a `btreeset` here.

src/parser.rs
```diff
@@ -529,10 +531,15 @@ fn get_xml_attribute<R: BufRead>(
     )))
 }
 
+struct JacocoReport {
```
calixteman, line 534: nit: we use to have this kind of stuff in defs.rs.

src/parser.rs
```diff
@@ -865,11 +873,11 @@ mod tests {
         assert!(result.functions.contains_key("MainProcessSingleton"));
         let func = result.functions.get("MainProcessSingleton").unwrap();
         assert_eq!(func.start, 15);
-        assert_eq!(func.executed, true);
+        assert!(func.executed);
```
calixteman, line 876: We'll lose the benefit of using `pretty_assertions` here, but do we care ?? @marco-c what's your opinion ?

calixteman (approved):
All in all, looks good for me with all tests green.
Thanks a lot for doing that.

src/parser.rs
marco-c: I guess they're very few cases, so we can live with it if it's what clippy wants

src/output.rs
dnaka91: I don't have any benchmarks for that, mostly just followed clippy's recommendations.
It probably doesn't make a difference in this case and just satisfies Google 😅

src/parser.rs
dnaka91: Alternatively I can simply turn of that specific clippy lint if that's more preferable.

src/parser.rs
dnaka91: I moved it over to defs.rs 🙇

dnaka91:
I have updated the PR to fix the windows build errors and added a few fixes for lints coming from the nightly version of clippy.
Anything else to do for this to be merged @calixteman?

I have a few more changes piled up and based on this PR. Would be lovely if this could be merged soonish 🙇‍♂️

marco-c:
@dnaka91 I think we're good, could you rebase on top of current master?

dnaka91:
@marco-c done, just rebased on top of master 🙇

marco-c (approved):

PR #633, opened 2021-06-07
Update semver to 1.0

dnaka91:
Bump semver to the latest release with minor code changes to how Version instances are created.
This is basically the same as #630 but with the necessary changes that dependabot can't do on its
own.


marco-c (approved):
Thanks!

PR #640, opened 2021-06-14
Fix cobertura report for multiple files

### LINKED ISSUE #636: Cobertura report has wrong summary of lines coverage
We started to use cobertura reports for code coverage in a project on GitLab: https://gitlab.com/svenfoo/sparschwein

The numbers reported on the `coverage` element do not seem to make much sense though:

     <coverage lines-covered="103" lines-valid="116"  ... >

There are certainly more than 116 lines in the code base. The XML lists several packages, each with lines that should sum up to something like 750 lines (counted with `wc -l` so that number might be off, but you get the point).

I had a look at the code in `cobertura.rs` and it seems to collect the total number of lines, and the number of covered lines, by calling `self.packages.get_lines()`. That is implemented as a Trait on `Vec<T>` and as far as I understand the implementation it pushes the lines of all packages into a single `FxHashMap` using the line number as the key. Is that right? 

If my interpretation of the code is right, then lines with the same number would be handled as the same line. That would then explain the result, as lines from different files (packages) would collide, leaving only about the number of lines in the largest package in the resulting hashmap.

Of course my interpretation of the code might be totally wrong. But still the reported numbers do not seem to reflect the number of lines in the code base. I got skeptical when the reported line coverage went down after I added more tests.

I am attaching the cobertura report here:
[coverage.xml.gz](https://github.com/mozilla/grcov/files/6626699/coverage.xml.gz)

svenfoo:
Fix the implementation of `Coverage::get_stats()` so that it doesn't create it from a merged hash of all lines from all packages, but instead adds up the the `Stats` of all packages.

Fixes #636

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -33,12 +33,27 @@ struct Coverage {
     packages: Vec<Package>,
 }
 
+#[derive(Default)]
 struct CoverageStats {
     lines_covered: f64,
     lines_valid: f64,
     branches_covered: f64,
     branches_valid: f64,
     complexity: f64,
+}
+
+impl std::ops::Add for CoverageStats {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        Self {
+            lines_covered: self.lines_covered + rhs.lines_covered,
+            lines_valid: self.lines_valid + rhs.lines_valid,
+            branches_covered: self.branches_covered + rhs.branches_covered,
+            branches_valid: self.branches_valid + rhs.branches_valid,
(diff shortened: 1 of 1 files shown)


src/cobertura.rs
```diff
@@ -41,6 +42,20 @@ struct CoverageStats {
     complexity: f64,
 }
 
+impl std::ops::Add for CoverageStats {
```
svenfoo, line 45: This could be derived using https://crates.io/crates/derive_more. Just let me know if you would prefer that.

marco-c (approved):
Thanks!

PR #643, opened 2021-06-14
Use Event::Text to write the source path to the cobertura report

svenfoo:
This removes an unwanted line-break from the <source> element's value.

Without this change the XML looks like

     <sources>
         <source>.
         </source>
    </sources>

This is changed to

     <sources>
         <source>.</source>
    </sources>

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -386,7 +386,9 @@ pub fn output_cobertura(results: CovResultIter, output_file: Option<&str>, deman
                 source_tag.len(),
             )))
             .unwrap();
-        writer.write(path.as_bytes()).unwrap();
+        writer
+            .write_event(Event::Text(BytesText::from_plain_str(path)))
+            .unwrap();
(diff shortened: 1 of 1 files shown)


marco-c (approved):
LGTM, needs a rebase

src/cobertura.rs
```diff
@@ -786,7 +792,7 @@ mod tests {
 
         let results = read_file(&file_path);
 
-        println!("{}\n", results);
```
marco-c, line 789: Nit: please keep the print, it can be useful for debugging

svenfoo: This line actually slipped in accidentally with commit 8048843df2ebc08c0f56d7e12da56d8519f25b94. No other tests output the resulting XML to stdout.

Of course if you want me to keep it, I can do that. But then I'd prefer to add it for all tests.

marco-c (approved):

PR #646, opened 2021-06-18
Remove remaining pre-2018 edition code

dnaka91:
Although Rust 2018 edition is defined in the Cargo.toml, there were still many code parts that used 2015 style.

Adding the `rust_2018_idioms` lint helped in identifying the remaining 2015 parts and I converted them to 2018 style. That is mostly:
- Remove `extern crate` statements.
- Add hidden lifetimes like `std::fmt::Formatter<'_>` instead of just `std::fmt::Formatter`.
- Use external macros directly without `#[macro_use]`.

In addition, I noticed that the `tcmalloc` crate doesn't compile on Windows (discovered while cross compiling from MacOS). Therefore, I further limited the `tc` feature to only work on unix platforms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,24 +1,6 @@
 #![recursion_limit = "1024"]
+#![deny(rust_2018_idioms, clippy::all)]
 #![allow(clippy::too_many_arguments)]
-
-extern crate cargo_binutils;
-extern crate chrono;
-extern crate crossbeam;
-extern crate fomat_macros;
-extern crate globset;
-#[macro_use]
-extern crate lazy_static;
-extern crate log;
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -1,25 +1,7 @@
 #![recursion_limit = "1024"]
+#![deny(rust_2018_idioms, clippy::all)]
```
marco-c, line 2: Instead of defining it in the code itself, let's add the option to the clippy call in the pre-commit configuration.

dnaka91: Done. I also removed the `clippy::all` again as that's the default anyways iirc

marco-c (approved):

PR #647, opened 2021-06-18
Fix all benchmarks

dnaka91:
I noticed that benchmarks currently don't compile and fixed them accordingly. Not 100% what each benchmark does so hopefully there is no difference in the benchmarks compared to what they did when they still worked.

Also, I had to make the `LittleEndian` & `BigEndian` structs public for the benchmarks. That two structs are likely helpful for the API in general as well, as the trait is exposed but until now no default implementations for it.

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -1,7 +1,5 @@
 #![feature(test)]
-extern crate crossbeam;
-extern crate grcov;
-extern crate rustc_hash;
+#![allow(clippy::unit_arg)]
 extern crate test;
 
 use crossbeam::channel::unbounded;
@@ -85,7 +83,6 @@ fn bench_lib_consumer(b: &mut Bencher) {
         FxHashMap::with_capacity_and_hasher(20_000, Default::default()),
     ));
     let (sender, receiver) = unbounded();
-    let source_root = None;
     let working_dir = PathBuf::from("");
     let gcno_buf: Vec<u8> = vec![
         111, 110, 99, 103, 42, 50, 48, 52, 74, 200, 254, 66, 0, 0, 0, 1, 9, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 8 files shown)


marco-c:
Could you add a command to build the benchmarks (without running them) as part of CI? This way at least we won't regress building them.

dnaka91:
That's a good idea thanks. I just pushed some changes to add an extra build step. Not sure whether this actually works on AppVeyor though, as the nightly version currently always fails.

Also, since the move from travis-ci.org to travis-ci.com it seems that GitHub doesn't pick up the status in the PRs anymore. So if something went wrong with the current change to the CI scripts, I won't be able to see it 🤔

benches/lib.rs
```diff
@@ -127,7 +123,7 @@ fn bench_lib_consumer(b: &mut Bencher) {
         for _ in 0..10_000 {
             sender
                 .send(Some(WorkItem {
-                    format: ItemFormat::GCNO,
+                    format: ItemFormat::Gcno,
```
marco-c, line 126: This should be GCNO

benches/reader.rs
```diff
@@ -1,37 +1,47 @@
 #![feature(test)]
-extern crate grcov;
+#![allow(clippy::unit_arg)]
 extern crate test;
 
-use grcov::{GcovReaderBuf, GCNO};
+use grcov::{Gcno, GcovReaderBuf, LittleEndian};
```
marco-c, line 5: This should be GCNO

marco-c:
> So if something went wrong with the current change to the CI scripts, I won't be able to see it

True, but at least we'll be able to see the results at some point when we fix CI (by moving away from Travis) :P

benches/lib.rs
dnaka91: Actually not anymore after I fixed the clippy warnings back in the other PR.

As clippy recommends writing even acronyms in Pascal case/camel case. Like `HTML` becoming `Html` and `TCP` being `Tcp`.

We can if course roll that back but that would probably be better fit as a separate PR.

marco-c (approved):

PR #648, opened 2021-06-18
Replace clap with structopt

dnaka91:
This replaces the manual setup of argument parsing through clap with structopt. It makes input handling in my opinion more simple and avoids the mismatch between argument definition and retrieval as structopt makes sure that all input parameters can be converted into the defined types.

I diffed the old help output with the new one to make sure that the differences are kept to a minimum.

While implementing the changes I potentially found a bug in regards to the `path_mapping` argument. It is defined as an argument that takes **1 or more** values, thus allowing to give multiple values at once. Later it is parsed as a **single value** only though. In this new version it correctly allows **only one** value as input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -484,7 +484,6 @@ version = "0.8.0"
 dependencies = [
  "cargo-binutils",
  "chrono",
- "clap",
  "crossbeam",
  "flate2",
  "fomat-macros",
@@ -505,6 +504,7 @@ dependencies = [
  "serde_json",
  "simplelog",
  "smallvec",
+ "structopt",
  "symbolic-common",
  "symbolic-demangle",
  "tcmalloc",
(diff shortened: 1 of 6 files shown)


marco-c:
The handling of the parameters is indeed simpler this way, but according to https://github.com/clap-rs/clap/issues/1574#issuecomment-575919336 structopt is going to be deprecated in favor of clap_derive.
Maybe we should hold off on this until the situation is clearer?

dnaka91:
I've used `structopt` as well as the current `clap` 3.0.0-beta.2 which includes the derive feature directly in clap.
Both are almost identical because they copied most of the code from structopt. Switching forth and back between them is therefore very easy and a future upgrade to clap 3.0 should be a quick change. Most of it is just switching the keywords for the derive options.

Besides that, between claps beta 1 and 2 a long time passed by, so it'll probably still take quite a long while until the full 3.0 is released.

With all that said I would recommend using structopt now. It's unclear when clap 3.0 will be out and the upgrade from structopt to clap is very easy (which I would happily do in the future).

calixteman (approved):
I don't see any obvious errors and overall it looks better with structopt.
So r+.

marco-c (approved):

marco-c:
> Thanks for the merge heart

Thank you for the contribution :wink:

PR #649, opened 2021-06-18
Include 'source_dir' in cobertura report

svenfoo:
If a source_dir is specified on the command-line, then all files in the
cobertura report are specified relative to that folder. The report should
then mention that source_dir instead of '.' so that files are found for
coverage display.

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -232,10 +232,10 @@ impl ToString for ConditionType {
 
 fn get_coverage(
     results: CovResultIter,
+    sources: Vec<String>,
     demangle: bool,
     demangle_options: DemangleOptions,
 ) -> Coverage {
-    let sources = vec![".".to_owned()];
     let packages: Vec<Package> = results
         .map(|(_, rel_path, result)| {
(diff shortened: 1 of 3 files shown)


src/cobertura.rs
```diff
@@ -232,14 +232,13 @@ impl ToString for ConditionType {
...
     results: CovResultIter,
+    sources: Vec<String>,
     demangle: bool,
     demangle_options: DemangleOptions,
 ) -> Coverage {
-    let sources = vec![".".to_owned()];
     let packages: Vec<Package> = results
         .map(|(_, rel_path, result)| {
             let all_lines: Vec<u32> = result.lines.iter().map(|(k, _)| k).cloned().collect();
-
```
marco-c, line 242: Nit: please reinstate the newline, as it is an unrelated change

marco-c (approved):
Thanks!

src/cobertura.rs
svenfoo: No problem. Done.

marco-c (approved):

PR #655, opened 2021-07-01
Update Documentation related to codecov

### LINKED ISSUE #632: Feature Request: Add a codecov output format
Codecov supports partial lines in its own format, but when using the coveralls format to upload to codecov, this granularity is lost.

It should be possible to determine for a line with multiple branches present how many of those branches were hit, which is exactly what the format captures.

specification of the format:
https://docs.codecov.io/docs/codecov-custom-coverage-format

maboesanman:
codecov works better with lcov than with coveralls format.

closes #632

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,9 @@ By passing `-t lcov` you could generate an lcov.info file and pass it to genhtml
 genhtml -o ./target/debug/coverage/ --show-details --highlight --ignore-errors source --legend ./target/debug/lcov.info
 ```
 
-#### Coveralls/Codecov output
+LCOV output should be used when uploading to Codecov, with the `--branch` argument for partial coverage support.
+
+#### Coveralls output
 
 Coverage can also be generated in coveralls format:
 
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ By passing `-t lcov` you could generate an lcov.info file and pass it to genhtml
 genhtml -o ./target/debug/coverage/ --show-details --highlight --ignore-errors source --legend ./target/debug/lcov.info
 ```
 
-LCOV output should be used when uploading to Codecov, with the `--branch` argument for partial line coverage support.
+LCOV output should be used when uploading to Codecov, with the `--branch` argument for branch coverage support.
 
 #### Coveralls output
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -239,7 +239,9 @@ By passing `-t lcov` you could generate an lcov.info file and pass it to genhtml
 genhtml -o ./target/debug/coverage/ --show-details --highlight --ignore-errors source --legend ./target/debug/lcov.info
 ```
 
-#### Coveralls/Codecov output
+LCOV output should be used when uploading to Codecov, with the `--branch` argument for partial coverage support.
```
marco-c, line 242: Wdym by "for partial coverage support"?

maboesanman: Partial coverage means the line is only partially covered. For example if you have an if statement which is short circuited:

`if(a | b)` would only be partially covered if the tests only ever run it for `a == true`

Do you have any suggestions for clearer wording?

marco-c: `partial coverage support` sounds like the coverage support is partial. That is, not fully implemented.

What about simply `branch coverage support`?

marco-c: Another option could be `to enable partial line coverage`.

maboesanman: That makes sense. Technically codecov displays if the line is partially covered, which is computed from branch coverage, but the effect is the same. I think you're right; just calling it branch coverage is clearer

README.md
```diff
@@ -239,7 +239,9 @@ By passing `-t lcov` you could generate an lcov.info file and pass it to genhtml
 genhtml -o ./target/debug/coverage/ --show-details --highlight --ignore-errors source --legend ./target/debug/lcov.info
 ```
 
-#### Coveralls/Codecov output
+LCOV output should be used when uploading to Codecov, with the `--branch` argument for partial line coverage support.
```
marco-c, line 242: I think this is still a little bit unclear, people might think the support is partial.
I'd go with either "with the `--branch` argument for branch coverage support" or "with the `--branch` argument to enable partial-line coverage".

marco-c (approved):

marco-c:
Thanks!

PR #661, opened 2021-07-19
Use newer macOS on Travis

wagenet:
I'd love to have macOS builds in the releases again. This does run, but the CI still fails.


marco-c:
We are planning to move off of Travis, though I'm not sure when we'll be able to.

marco-c:
We are no longer using Travis-CI.

Issue #637, opened 2021-06-14
travis-ci.org shutting down tomorrow on June 15th


dnaka91:
When checking the builds of my PR I noticed that travis-ci.ORG is shutting down tomorrow as they migrate everything to a single travis-ci.COM instead of having to separate domains (probably because it caused confusion in the past).

> Hi! It seems there are still plenty of builds running. We're pushing back the shutdown date to June 15th. Please consider migrating to travis-ci.com. 

Migration should be easy, just someone with contributor rights needs to link travis-ci.com as CI to this repo.

Besides that, would there be interest in migrating to GitHub Actions?
I'm quite familiar with it and could prepare the changes to have everything run there instead.

marco-c:
Our plan would actually be to switch to Taskcluster (#135), as we are using it for many other projects and it is super configurable.

marco-c:
I've moved the CI to travis-ci.com, so I'm going to close this.

If you want to contribute switching to Taskcluster, it'd be great.
Otherwise, I'd also be OK with switching to GitHub Actions for the time being and switch to Taskcluster in the future.

dnaka91:
Hm, I just updated my PRs and it seems that GitHub is not receiving any status from travis-ci.com 🤔.

Regarding Taskcluster, I never worked with it so probably a bit difficult for me to prepare the initial move over. If there is already a WIP branch or similar I'd be glad to help though.

marco-c:
For taskcluster, you could take inspiration from https://github.com/mozilla/rust-code-analysis/blob/master/.taskcluster.yml.
Otherwise, we can try with GitHub Actions.

marco-c:
@dnaka91 still interested in this? If you know how to use GitHub Actions well, we can move to that instead of Taskcluster for the time being (still better than no CI on Linux and Mac!)

dnaka91:
I was looking at Taskcluster a bit but as I haven't used it myself yet, it would probably take me many iterations before we get anything useful out of that migration.

I'll start on setting up everything as GitHub Actions for now then. Shouldn't be too complicated except the compiler setups eventually.

Have to figure out how to test the GitHub Actions before merging as I'm not a collaborator of the repo. Maybe I can get it running on my fork at first and then you'd have to enable Actions on this main repo once I open the PR.

marco-c:
[...]

PR #667, opened 2021-07-31
Configure GitHub Actions

dnaka91:
Configuring GitHub Actions as CI solution for now, as discussed in #637.
Maybe @marco-c  could answer me a few open questions (or anybody else who knows about the current CI setup 😄).

Tested the setup on my fork and results can be seen [here](https://github.com/dnaka91/grcov/actions).

## Todo

- [ ] Run code coverage on grcov itself and upload.
- [ ] Automatically add pre-compiled binaries to new releases.

## Questions

- [x] What is the reason behind the test matrix on Travis CI? From my understanding it...
  - runs code coverage several times over grcov with different combinations of g++ and clang.
  - doesn't do anything specific with the configured compilers.
- [ ] Current _full_ test suite seems broken?
  - Full test suite in the sense of running `cargo test` instead of `cargo test --lib` to trigger everything that's under the `tests` folder.
  - The `Dockerfile` setup doesn't seem to work neither and always fails.
  - What is the solution to this? Is it some change in the compilers, some code change in `grcov` or something else?


marco-c:
>  What is the reason behind the test matrix on Travis CI?

We need to make sure grcov is able to parse coverage artifacts from all currently supported versions of the compilers.

>  Current full test suite seems broken? 

What exactly is broken?
Let's run it anyways, we can fix it in follow-up PRs.

marco-c:
@dnaka91 btw, GitHub Actions is enabled for this repo, so no need to test this PR on your fork. You will need to only use GitHub official actions though, and so stop using pre-commit/action (which is deprecated anyway) and the action-rs ones, as they are currently not allowed to run on mozilla repositories. It should be trivial to manually run the commands they implement.

dnaka91:
Thanks for the quick answers.

Too bad we're not allowed to run 3rd party GitHub Actions. No big trouble for the current setup though, just a few more manual steps instead.

> We need to make sure grcov is able to parse coverage artifacts from all currently supported versions of the compilers.

Okay, got it. I saw that there are several coverage outputs in the `tests/` folder and thought they would suffice (also keeping build times down). Maybe in the future it would be better to add a small extra Rust program for that purpose instead of compiling grcov which is rather heavy.

Anyways, for now I will prepare the steps to test grcov against itself, as it is in the old tests.

> What exactly is broken?

Several integration tests fail. That can be easily reproduced by just building the `tests/Dockerfile`.

> [...] no need to test this PR on your fork [...]

I think that actually happens automatically, so it will build on the fork as it sees the Github Actions settings there too. Well, I guess I can disable it now that I know they're enabled here too 👍

marco-c:
> Okay, got it. I saw that there are several coverage outputs in the tests/ folder and thought they would suffice (also keeping build times down). Maybe in the future it would be better to add a small extra Rust program for that purpose instead of compiling grcov which is rather heavy.

> Anyways, for now I will prepare the steps to test grcov against itself, as it is in the old tests.

We are not building grcov against itself, but building some small programs and running them, then grcov is used to parse the coverage artifacts coming from those small programs.
We need to test building those small programs with multiple compilers (that's why we have the matrix, but actually we could also have just one job with multiple compilers installed and serially compile the programs with the different compilers, or even better a job that builds grcov and multiple jobs, one for each compiler, that gets grcov from the first job and runs it on the compiled programs).

marco-c:
> Several integration tests fail. That can be easily reproduced by just building the tests/Dockerfile.

I've opened https://github.com/mozilla/grcov/issues/672 for this.

dnaka91:
Thanks for the quick answers as always 🙇 .
Still, I don't see anything in the current Travis CI setup that builds some custom programs to generate coverage data with.

```yaml
- |
  if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then
    export RUSTFLAGS="-Zinstrument-coverage";
  fi
- if [ "$TRAVIS_TAG" ]; then export CARGO_OPTIONS="--release"; fi
- cargo build --verbose $CARGO_OPTIONS
- LLVM_PROFILE_FILE="grcov-%p-%m.profraw" cargo test --verbose $CARGO_OPTIONS
- |
  if [ "$TRAVIS_RUST_VERSION" = "nightly" ] && [ -z "$TRAVIS_TAG" ]; then
    ./target/debug/grcov `find . \( -name "grcov-*.profraw" \) -print` --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" -o lcov.info;
    bash <(curl -s https://codecov.io/bash) -f lcov.info;
  fi
```

So all the `if` statements aside I understand that it does the following:
- Build grcov.
- Run grcov's tests with code coverage.
- Execute previously built grcov to turn `grcov-*profraw` coverage files into `lcov.info` coverage data.
- Upload to codecov.io.

All other parts of the CI focus around running `pre-commit` building release binaries and building benchmark programs. So I don't really see where the env vars for `gcc` and `clang` come into play. I just tried it out to see whether they affect Rust's compilation process in any way, but it seems Rust doesn't care about those https://github.com/dnaka91/grcov/actions/runs/1096729282.

The only real difference here is that I run `cargo test --lib` instead of `cargo test` because the integration tests don't work at the moment. Are these tests internally building the extra programs and generate their coverage data?

dnaka91:
Okay, I think I'll wait for #672 to be solved as I can't properly test the matrix setup with all these tests failing.
I thought of just adjusting the coverage files until tests pass but it's just too many.

Looking at these tests the tool [insta](https://github.com/mitsuhiko/insta) might be a nice helper to confirm the right output. But that's up to another PR and I'll maybe create an issue for that in the future to further discuss it.

Currently Windows tests seem to fail due to the `llvm-tools-preview` not being compiled with zlib support. Not sure what to do about that one except maybe raising an issue on the rust repo itself 🤔

marco-c:
> Okay, I think I'll wait for #672 to be solved as I can't properly test the matrix setup with all these tests failing.

#672 might take a while to be fixed. We can start by landing a workflow that doesn't run the integration tests (so just running `cargo test --lib`).

marco-c:
> Looking at these tests the tool insta might be a nice helper to confirm the right output. But that's up to another PR and I'll maybe create an issue for that in the future to further discuss it.

Sounds interesting, definitely worth investigating.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
+name: CI
```
marco-c, line 1: Let's keep the workflow in the CICD.yml file

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+          rustup override set ${{ matrix.toolchain }}
+      - name: Configure cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
```
marco-c, line 26: Why caching `target/`? So we can do incremental compilation?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+          key: test-${{ runner.os }}-cargo-${{ matrix.toolchain }}-${{ hashFiles('**/Cargo.lock') }}
+      - name: Test
+        run: cargo test --lib
+  lint:
```
marco-c, line 30: There is some redundancy between `lint` and `pre-commit`. `pre-commit` is running rustfmt and clippy too.
Let's remove the `lint` entry and just use `pre-commit` for linting (but rename the `pre-commit` entry to `lint`, since the fact we're using `pre-commit` for linting is just an "implementation detail" which might change).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+      - name: Setup Rust
+        run: |
+          rustup update stable
+          rustup override set stable
+      - name: Install pre-commit
+        run: pip install pre-commit
```
marco-c, line 67: If we are not caching the pip dir, use the `--no-cache-dir` option

src/parser.rs
```diff
@@ -79,7 +79,7 @@ macro_rules! try_next {
 
 macro_rules! try_parse_next {
     ($v:expr, $l:expr) => {
-        try_parse!(try_next!($v, $l), $l);
```
marco-c, line 82: Let's make these changes in separate PRs.

.github/workflows/ci.yml
dnaka91: Didn't know about that flag, good point 👍

.github/workflows/ci.yml
dnaka91: There definitely is. I'll remove the lint action and rename the pre-commit job to lint.

Originally I added the lint as I copied the config from one of my own config and then adjusted it.

.github/workflows/ci.yml
dnaka91: Pretty much yes. It allows to run the test much sooner as we don't have to re-compile all the dependencies over and over again.

.github/workflows/ci.yml
dnaka91: Hm, my plan was to later add a separate `deploy.yml` workflow with the compile steps for release binaries. Wanted to add them as separate workflow to not have one huge setting but rather smaller purpose built files.

Would you like to have everything in one file instead?

marco-c: Yes for now I'd prefer a single file. We can split it later if we find it is becoming too large and hard to read.

.github/workflows/ci.yml
marco-c: It could be problematic for coverage calculation. We need to ensure we save coverage artifacts outside that directory

marco-c:
@dnaka91 still interested in finishing this up?

marco-c:
@dnaka91 we could merge this as a first iteration and work on fixing the failures later, wdyt?

marco-c:
@dnaka91 ping

dnaka91:
Hey @marco-c, currently I'm just waiting on the tests being fixed. We can of course merge this first, but it doesn't really give any benefit if the tests fail every time anyways right?

marco-c:
@dnaka91 I've fixed the integration tests for that Dockerfile (basically, I've fixed them for Clang 7).

We should make sure we support LLVM/GCC from Debian stable, and all versions after those.
So, in the test matrix, we'll need to test the same Clang and GCC versions we had in the .travis.yml plus the new ones that were released since then.

dnaka91:
Got it. I merged the latest changes from master, hopefully tests work now. If all lights are green I'll add back in the other LLVM/GCC combinations from the Travis CI config.

marco-c:
@dnaka91 I filed https://github.com/mozilla/grcov/issues/727 for the remaining failing test and just fixed it on master.
If you rebase again, things should be green.

dnaka91:
It's a never ending loop of issues 😅. I remember this from before, somehow on Windows the zlib compression support is missing from the tools of `llvm-preview` it seems?

Not sure how to solve that one to be honest. Maybe in general it would be better to pin to a specific nightly version as there can be new breaking changes in nightly any day?

marco-c:
The solution would be to build a simple program and generate the profile as part of the test, so the machine which generates it is also the machine which reads it.

We don't have to solve everything in order to land this though, it's OK if some tests are broken and we deal with them as follow-ups.

> Maybe in general it would be better to pin to a specific nightly version as there can be new breaking changes in nightly any day?

We want to know as soon as a new Nightly breaks so we can fix it. We'll have to accept the fact that it can break any day.

src/llvm_tools.rs
```diff
@@ -110,7 +110,6 @@ mod tests {
             &PathBuf::from("src"),
             &tmp_path,
         );
-        assert!(lcovs.is_ok());
```
marco-c, line 113: Can you bring these checks back?

marco-c:
> The solution would be to build a simple program and generate the profile as part of the test, so the machine which generates it is also the machine which reads it.

There must be some other solution as Appveyor is working... Though ^ is the best solution long term, so let's not worry about that.

src/llvm_tools.rs
dnaka91: Sure thing 👍

.github/workflows/CICD.yml
```diff
@@ -1,18 +1,61 @@
...
-  build:
-
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        toolchain: [stable, nightly]
+    continue-on-error: ${{ matrix.os == 'windows-latest' && matrix.toolchain == 'nightly' }}
```
marco-c, line 11: Is it possible to always continue on errors?

dnaka91: Sure no problem. 

According to the docs, that would mean a PR is valid even in case of an error, though. So, if there is a real error in one of the other test cases, it would be considered okay for merge.

If that side effect is okay with you, I'll mark all tests to continue on error :+1:

marco-c:
Yes, that's fine, thanks!

marco-c:
Let's use the versions of GCC and Clang from Debian stable as the minimum we support. So GCC 10 (https://packages.debian.org/bullseye/gcc) and Clang 11 (https://packages.debian.org/bullseye/clang).
Then, we test all versions starting from those up to the most recent ones (GCC 11 and Clang 13).

.github/workflows/CICD.yml
```diff
@@ -1,18 +1,148 @@
...
+          key: test-${{ runner.os }}-cargo-${{ matrix.toolchain }}-${{ hashFiles('**/Cargo.lock') }}
+      - name: Test
+        run: cargo test --lib
+  test-coverage:
+    name: Test coverage (gcc-${{ matrix.gcc }}, clang-${{ matrix.clang }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - gcc: 5
```
calixteman, line 37: @marco-c I think we could remove gcc 5-6, gcc version in ubuntu 18.04 is 7.4.

.github/workflows/CICD.yml
```diff
@@ -1,18 +1,148 @@
...
+            deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main
+            deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main
+            deb http://apt.llvm.org/focal/ llvm-toolchain-focal-12 main
+            deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main
+          EOF
+          sudo mv /tmp/extra.list /etc/apt/sources.list.d/
 
+          # Install toolchains
+          sudo apt-get update
+          sudo apt-get install -y clang-${{ matrix.clang }} g++-${{ matrix.gcc }}
```
calixteman, line 82: Disclaimer: I don't know anything about github actions, so in advance, sorry for the stupid questions.

Are those packages cached at some point ? If not, is it possible to cache them ?

dnaka91: There is a caching action to do this, but we have to manually define the folders and files to cache.

Currently, it's already set up for Rust but not for APT. I found some community made actions that automate caching of APT, but according to @marco-c they're not permitted in Mozilla repos.

I'm not an expert on APT's structure, but a quick search yielded that the repo lists are cached at `/var/lib/apt/lists` and the DEBs are cached at `/var/cache/apt/archives`.

We might just give it a try and throw these in a GitHub Actions cache. Just not sure if that's enough, as `/var/cache/apt` contains some extra cache files, and I'm not sure about their relation to the cached DEBs.

Also, if we cache these files, how do we evict the cache? For Cargo, it's simple as we invalidate the cache whenever the `Cargo.lock` file changes, but for APT it seems more difficult as we don't have a lock file.

.github/workflows/CICD.yml
dnaka91: In addition to that, should I add newer combinations like clang 12 and 13 as well?

dnaka91:
@marco-c about the build failing for Windows with nightly in GitHub Actions but working in AppVeyor.

I see that the AppVeyor config installs LLVM from [here](https://github.com/marco-c/llvm-appveyor/releases/tag/llvm_342383) instead of using an official build. That might be the issue for the build problems in GH Actions?

If that might be the case, I can try doing the same installation step in GH Actions as well. But my concern with that is, shouldn't it just work against the normal LLVM builds or at least an official pre-compiled package? Sounds like it's bound to fail on a typical dev machine as well, as I would think the default builds are more likely installed 🤔.

Maybe it's worth installing LLVM on our own, but from the official installers (for example through **Chocolatey**, which is pre-installed on GH Actions runners), instead of relying on the pre-installed version.

.github/workflows/CICD.yml
marco-c: Let's file this as a follow-up.

.github/workflows/CICD.yml
marco-c: Yes, see https://github.com/mozilla/grcov/pull/667#issuecomment-985666987.

marco-c:
> If that might be the case, I can try doing the same installation step in GH Actions as well. But my concern with that is, shouldn't it just work against the normal LLVM builds or at least an official pre-compiled package? Sounds like it's bound to fail on a typical dev machine as well, as I would think the default builds are more likely installed thinking.

Yes, it was kind of a requirement in appveyor, but we shouldn't do it.
Let's file the Windows error as a follow-up issue to investigate. I don't want to block this on Windows support.

.github/workflows/CICD.yml
```diff
@@ -1,18 +1,124 @@
...
+          key: test-${{ runner.os }}-cargo-${{ matrix.toolchain }}-${{ hashFiles('**/Cargo.lock') }}
+      - name: Test
+        run: cargo test --lib
+  test-coverage:
+    name: Test coverage (gcc-${{ matrix.gcc }}, clang-${{ matrix.clang }})
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        gcc: [10, 11]
+        clang: [11, 12, 13]
```
marco-c, line 37: There's no need to test all combinations, we just need to make sure there is at least one task for each gcc version and each clang version.

E.g.:
- gcc 10, clang 11
- gcc 11, clang 12
- gcc 10, clang 13

Instead of:
- gcc 10, clang 11
- gcc 11, clang 11
- gcc 10, clang 12
- gcc 11, clang 12
- gcc 10, clang 13
- gcc 11, clang 13

marco-c:
This is looking better and better, I think these are the only things left to do:
- implement coverage uploading
- publish releases
- fix Windows failures
- fix GCC failures (probably just need specific "expected output" files)
- add caching for APT packages

We can merge this as-is and file the previous as follow-ups, unless @dnaka91 you have time to finish the coverage uploading and release publishing in the next few days.

dnaka91:
Uploading coverage should be easy to do.

Attaching artifacts to releases could be a bit more tricky, as GitHub deprecated their own official actions in favor of community actions the last time I checked. So will have to do a bit of manual hacking with `curl` or so as we can't use community actions.

Another thing I would like to add to the TODO is:
- Add matrix entries to the `Test coverage` job for MacOS and Windows. Currently missing in comparison to the Travis CI setup.

marco-c:
We can use some community actions, if they are hard to implement ourselves and widespread enough we can ask for exceptions to the rule.

dnaka91:
Okay, there are some alternatives directly linked from the outdated GitHub action. The two that look fitting are:
- <https://github.com/ncipollo/release-action>: Used this one several times successfully and has pretty much all you need.
- <https://github.com/softprops/action-gh-release>: Has a much higher star count compared to the other, and feature sets seem to be pretty similar. Just haven't used this so far.

Tomorrow I have quite some time, so I'll check out how complicated this gets if we try this without 3rd party actions.

One thing I'd like to confirm first though. Would you like it to work like:
- You create a release with title and description, then, once published, the action will trigger and attach the binaries afterwards.
- The action is trigger by pushing a new tag, compiles all binaries, then creates the release **and** attaches the binaries. Release text could be possibly generated/extracted from somewhere or added later by you and other maintainers.

marco-c:
> The action is trigger by pushing a new tag, compiles all binaries, then creates the release and attaches the binaries. Release text could be possibly generated/extracted from somewhere or added later by you and other maintainers.

I'd prefer this option.

dnaka91:
Okay, after many runs and hours of trial and error (mostly due to the cross-compilation part) I finally got a working “release on tag” step.

@marco-c I know you like to have everything in a single file, but in this case it was needed so I could apply a filter to the action to only run when the tag has the format `v*.*.*`.

I just pushed the changes, but you won't be able to see it working until this is merged and a new release is tagged. Therefore, I temporarily merged these changes on the master branch of my fork and made a random release. You can see the results here:
- action: https://github.com/dnaka91/grcov/actions/runs/1585796255
- release: https://github.com/dnaka91/grcov/releases/tag/v0.100.0

marco-c:
This is looking great, thanks for helping us!

I'm going to merge this as-is so we have a full working CI again, then we can take care of the rest in follow-up PRs.

marco-c (approved):

marco-c:
I've filed #734, #735, #736, #737, #738, #739, #740.

marco-c:
Note: using the target name will break people relying on the current naming. I guess that's an acceptable breakage.

I have filed #741 to add the tcmalloc build some users rely on (it's faster and uses less memory).

PR #674, opened 2021-08-04
Clean profraw files after running the integration test binaries, or grcov will think source-based coverage was used

### LINKED ISSUE #673: `[ERROR] The path to the compiled binary must be given as an argument when source-based coverage is used` when running integration tests
The integration tests are only testing gcov-based coverage at this time, so this error is fishy.

marco-c:
Fixes #673


calixteman (approved):
LGTM, thx.

PR #686, opened 2021-09-29
``--no-method`` option for Cobertura output

CurryPseudo:
I'm using [pycobertura](https://github.com/aconrad/pycobertura) as a code coverage diff tool, but grcov's cobertura output is not compatible with it.
pycobertura only accepet
```xml
<class><lines>...<lines><class>
```
but grcov 's cobertura output contains method infos like
```xml
<class><methods><method><lines>...<lines></method></methods></class>
``` 
This PR provide a option ``--no-method`` to generate the cobertura output without method info.

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -338,6 +338,7 @@ pub fn output_cobertura(
     results: CovResultIter,
     output_file: Option<&Path>,
     demangle: bool,
+    with_method: bool,
 ) {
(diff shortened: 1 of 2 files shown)


CurryPseudo:
@marco-c  not sure how to add a reviewer, help needed

marco-c:
@CurryPseudo did you try asking the pycobertura developers if they can support it?

PR #720, opened 2021-11-24
Fall back to copying file on Windows

sivadeilra:
Creating a symbolic link requires administrator privileges (by default)
on Windows. This commit falls back to copying a file if creating a
symbolic link fails, on Windows.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -3,7 +3,6 @@ use std::cell::RefCell;
 use std::env;
 use std::fs::{self, File};
 use std::io::{self, BufReader, Read};
-use std::os;
 use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
 use zip::ZipArchive;
@@ -280,16 +279,9 @@ impl Archive {
                 // don't use a hard link here because it can fail when src and dst are not on the same device
                 let src_path = dir.join(name);
 
-                #[cfg(unix)]
-                os::unix::fs::symlink(&src_path, path).unwrap_or_else(|_| {
+                crate::symlink::symlink_file(&src_path, path).unwrap_or_else(|_| {
                     panic!("Failed to create a symlink {:?} -> {:?}", src_path, path)
                 });
-
-                #[cfg(windows)]
-                os::windows::fs::symlink_file(&src_path, path).unwrap_or_else(|_| {
-                    panic!("Failed to create a symlink {:?} -> {:?}", src_path, path)
-                });
-
                 true
             }
             ArchiveType::Plain(_) => {
--- a/src/symlink.rs
+++ b/src/symlink.rs
@@ -1,0 +1,34 @@
+use std::path::Path;
+
+#[cfg(unix)]
+pub(crate) use std::os::unix::fs::symlink as symlink_file;
+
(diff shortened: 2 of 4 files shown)


src/producer.rs
```diff
@@ -285,11 +285,14 @@ impl Archive {
...
+                // rights. Fall back to copying if creating the symbolic link fails.
                 #[cfg(windows)]
-                os::windows::fs::symlink_file(&src_path, path).unwrap_or_else(|_| {
-                    panic!("Failed to create a symlink {:?} -> {:?}", src_path, path)
-                });
-
+                os::windows::fs::symlink_file(&src_path, &path)
+                    .or_else(|_| std::fs::copy(&src_path, &path).map(|_| ()))
+                    .unwrap_or_else(|_| {
+                        panic!("Failed to create a symlink {:?} -> {:?}", src_path, path)
```
marco-c, line 294: Could you update the error message?

marco-c:
Given that this is going to result in slowdown, maybe we should warn the user (only once) when we are not able to create symlinks and suggest them to run grcov with admin privileges.

src/producer.rs
sivadeilra: Sure. What would you like it to be?  Something like `"Failed to create symlink / copy file"` or `"Failed to copy file"`, or something else?

marco-c: If we implement the symlink creation warning I suggested below, we can say "Failed to copy file".
If we don't implement the warning, we can say "Failed to create symlink or copy file".

calixteman: It's a bit hard... with big projects it can become awful.
I wonder if should use an option to allow that...
If some people are using grcov with a kind of big project and they aren't aware about that stuff then each time it'll run a forest will burn. And if we've an option (`--copy-when-no-symlink` or whatever) then the first run will fail because the option isn't used, and then with a good error message they can decide to copy or to allow symlink creations in configuring windows correctly.

sivadeilra:
I like the idea of printing a warning. I've updated the PR. It will now print a warning, but only once per grcov. I added a section to `README.md` explaining how to enable symbolic links, and the warning should point people to that text.

src/symlink.rs
```diff
@@ -0,0 +1,34 @@
...
+/// On Windows, creating a symbolic link to a file requires Administrator
+/// rights. Fall back to copying if creating the symbolic link fails.
+#[cfg(windows)]
+pub(crate) fn symlink_file<P: AsRef<Path>, Q: AsRef<Path>>(
+    original: P,
+    link: Q,
+) -> std::io::Result<()> {
+    use std::sync::atomic::{AtomicBool, Ordering::SeqCst};
+
+    static HAVE_PRINTED_WARNING: AtomicBool = AtomicBool::new(false);
```
marco-c, line 17: An AtomicBool shouldn't be necessary because the producer is single threaded, but better to be safe than sorry in case we change that in the future.

marco-c (approved):

calixteman (approved):
LGTM.
Thanks for doing this: it's valuable (especially because we are linux devs).

Issue #675, opened 2021-08-15
attempt to divide by zero


skoky:
I am getting this issue

`08:10:42 [ERROR] A panic occurred at /Users/skokan/.cargo/registry/src/github.com-1ecc6299db9ec823/grcov-0.8.2/src/html.rs:458: attempt to divide by zero`

while runining

`grcov . --binary-path ./target/debug/ -s . -t html --branch --ignore-not-existing -o ./coverage/`

on my project with grcov 0.82.0
I follow steps described here https://github.com/marco-c/rust-code-coverage-sample

youduda:
After switching to rust-nighly as the default toolchain, did you run `rustup component add llvm-tools-preview`?

skoky:
Yes I did

ralpha:
For me the problem was solved after making sure the `llvm-tools-preview` was added in the nightly toolchain, not just stable.
`rustup toolchain install nightly --allow-downgrade --component llvm-tools-preview`

bausano:
I have the same issue with following setup:

```
active toolchain
----------------

nightly-x86_64-unknown-linux-gnu
rustc 1.58.0-nightly (dd549dcab 2021-11-25)
```

Then I add `llvm-tools-preview` making sure I am on nightly. Then I run

```
cargo test --lib

grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
```

And I get the error:

```
... many similar errors as these:
profiling: xxx-f9cc652c76a93eba.gcda: cannot merge previous GCDA file: corrupt arc tag (0x00000000)
profiling: xxx-f9cc652c76a93eba.gcda: cannot merge previous GCDA file: mismatched number of counters (6)
profiling: xxx-f9cc652c76a93eba.gcda: cannot merge previous GCDA file: corrupt arc tag (0x00000000)
profiling: xxx-f9cc652c76a93eba.gcda: cannot merge previous GCDA file: mismatched number of counters (6)
profiling: xxx-f9cc652c76a93eba.gcda: cannot merge previous run count: corrupt object tag (0x00000000
[...]

PR #721, opened 2021-11-24
Avoid divide-by-zero when total_lines == 0

sivadeilra:
If you run a report that matches no source files, then currently
grcov will divide by zero. This does the minimum necessary to avoid
that, by treating 0/0 as 0.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -455,7 +455,14 @@ pub fn gen_badge(tera: &Tera, stats: &HtmlStats, conf: &Config, output: &Path, s
     };
 
     let mut ctx = make_context();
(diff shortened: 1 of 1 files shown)


marco-c (approved):

marco-c:
Thanks!

jonhoo:
@marco-c Do you have a rough estimate on when the next grcov release might be? I'd really like this fix, but 0.8.2 is from July and so doesn't include it.

jonhoo:
Also, for reference, this fixes https://github.com/mozilla/grcov/issues/675.

marco-c:
@jonhoo I've released v0.8.5 with the fix.

PR #722, opened 2021-11-24
Fix compiler warning about `;` in macro

sivadeilra:
Rustc now warns about a terminal `;` in macros that are evaluated
in expression contexts. The fix is trivial -- remove the `;`.


marco-c (approved):

marco-c:
Thanks!

PR #724, opened 2021-11-26
Return 100% in case of empty file

calixteman:
- 0% means that something is wrong
- nothing is wrong with an empty file: it's just empty

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -6,6 +6,7 @@ use std::collections::HashMap;
 use std::collections::{btree_map, BTreeMap};
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader, Write};
+use std::ops::{Div, Mul};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


src/llvm_tools.rs
```diff
@@ -84,7 +84,7 @@ fn get_profdata_path() -> Result<PathBuf, String> {
     if !path.exists() {
         Err(String::from(
             "We couldn't find llvm-profdata. Try installing the llvm-tools \
-            component with `rustup component add llvm-tools-preview`."
+            component with `rustup component add llvm-tools-preview`.",
```
marco-c, line 87: ?

marco-c (dismissed):
LGTM, but please remove the unrelated change in src/llvm_tools.rs

marco-c:
Could you rebase on top of master? This will make the fmt thing go away.

marco-c (approved):

PR #756, opened 2022-01-23
Add log-level flag

dwhjames:
- Preserve existing behavior, which was a statically defined log level of ‘ERROR’
- Add `--log-level <LEVEL>` flag to explicitly set the level filter to one of the 6 valid levels: `["OFF", "ERROR","WARN", "INFO", "DEBUG", "TRACE"]`


marco-c (approved):
Thanks!

PR #759, opened 2022-01-23
Add all lines to class coverage

### LINKED ISSUE #758: Cobertura report doesn't contain proper class line coverage
The cobertura report format introduced with #599 seems to [only report orphan lines on class level](https://github.com/mozilla/grcov/blob/master/src/cobertura.rs#L314).

While the cobertura format itself is not properly specified, it seems that pretty much all other test frameworks and reporters always add the full coverage data on class level and not just orphan lines. This can also be seen when looking at example reports created by cobertura itself, e.g. the [cobertura example report for the official Jenkins cobertura plugin](https://github.com/jenkinsci/cobertura-plugin/blob/master/src/test/resources/hudson/plugins/cobertura/coverage-with-data.xml). The lack of this data causes incompatibilities with reporting tools like [GitLab's cobertura artefact](https://docs.gitlab.com/ee/user/project/merge_requests/test_coverage_visualization.html): **[Example coverage not shown since empty like coverage on class level was reported](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/1/diffs)** 

When the logic is changed to always add the complete class level line coverage ([via following diff](https://github.com/fh1ch/grcov/commit/0a64a1b16d34af6eb80a240a16feed2a0fda1713)), the report on GitLab works just fine: **[Example coverage shown](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/2/diffs)**

This fact also becomes apparent when bypassing the cobertura reporter by creating an lcov report first and using the [`lcov_cobertura` python module](https://github.com/eriwen/lcov-to-cobertura-xml) to create the final cobertura report:

```sh
grcov . --source-dir . --binary-path target/debug/ --branch --output-type lcov --output-path coverage.lcov
pip3 install lcov_cobertura
python3 -m lcov_cobertura coverage.lcov
```

This report contains the full class level line coverage and is hence also properly parsed by tools like GitLab.

fh1ch:
This PR changes the cobertura report format which was introduced with #599 to always return the complete line coverage information on class level. This aligns the class level line coverage behaviour with other reporters like [cobertura itself](https://github.com/jenkinsci/cobertura-plugin/blob/master/src/test/resources/hudson/plugins/cobertura/coverage-with-data.xml#L39) (Java), [nyc/istanbul.js](https://istanbul.js.org/docs/advanced/alternative-reporters/#cobertura ) (JavaScript) or [Coverage.py](https://github.com/nedbat/coveragepy/blob/6.2/tests/gold/xml/y_xml_branch/coverage.xml#L14) (Python) and fixes the integrations into tools like GitLab (as verified by [the following example](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/2/diffs)).

I can also provide a follow-up MR, which adds a [GitLab coberatura example to the readme](https://github.com/mozilla/grcov/issues/468#issuecomment-823544526), if this MR goes into the right direction.

Closes #758

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -241,8 +241,6 @@ fn get_coverage(
         .map(|(_, rel_path, result)| {
             let all_lines: Vec<u32> = result.lines.iter().map(|(k, _)| k).cloned().collect();
 
-            let mut orphan_lines: BTreeSet<u32> = all_lines.iter().cloned().collect();
-
             let end: u32 = result.lines.keys().last().unwrap_or(&0) + 1;
 
             let mut start_indexes: Vec<u32> = Vec::new();
@@ -295,7 +293,6 @@ fn get_coverage(
                         .filter(|&&x| x >= function.start && x < func_end)
                     {
                         lines_in_function.push(*line);
-                        orphan_lines.remove(line);
                     }
 
                     let lines: Vec<Line> = lines_in_function
@@ -311,7 +308,7 @@ fn get_coverage(
                 })
                 .collect();
 
-            let lines: Vec<Line> = orphan_lines.into_iter().map(line_from_number).collect();
+            let lines: Vec<Line> = all_lines.into_iter().map(line_from_number).collect();
             let class = Class {
                 name: rel_path
                     .file_stem()


marco-c:
CC @geovie @svenstaro @svenfoo, any thoughts?

svenfoo:
I find it hard to judge the change without seeing an example. The code it touches is not straight-forward, changes can have unexpected side-effects, the cobertura output format is clearly under-specified and I am also pretty sure that the current implementation in `grcov`is not correct. I think we really need to have tests for the cobertura report. What I am imaging is an acceptance test featuring a small example project that we instrument and then generate a report for that is then compared to the report we expect to get.

snue:
Thanks @fh1ch! I can confirm this solves the actual GitLab UI highlighting for me as well.

@svenfoo 
> I find it hard to judge the change without seeing an example.

You  can check out the linked example merge requests: [not working](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/1) / [fixed](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/2)
Only the fixed one shows colored coverage annotations on the diff.

If you download the `run tests:cobertura` artifacts from the pipeline results, the difference is that [the broken one](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/jobs/1979822259/artifacts/download?file_type=cobertura) has an empty `<lines>` section on the `<class>` level, whereas [the fixed one](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/jobs/2005702040/artifacts/download?file_type=cobertura) lists all the lines (as other cobertura xml generators do). GitLab obviously expects the latter.

> The code it touches is not straight-forward, changes can have unexpected side-effects

Regarding the suspected original intention of that code, it keeps a list of "orphaned lines" and will only list those that are not an explicit part of any `<methods>` for listing at the class level instead. The change removes the whole concept of "orphaned lines" and takes the original list verbatim, again as other generators do. I approve of this change. To me it is a rather clean simplification and I cannot spot chances for side effects. Any "orphaned lines" will still be listed as they are a strict subset.

Of course explicit tests are favorable (not arguing against that), but this solves an actual issue right now. I would appreciate a timely fix in an official release.

fh1ch:
@svenfoo @snue thanks a lot for the feedback folks!

I've added the two cobertura reports of [the linked example](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/2) for conveniency's sake here as well: 

<details>

**Before:**

```xml
<?xml version="1.0"?>
<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>
<coverage lines-covered="11" lines-valid="16" line-rate="0.6875" branches-covered="0" branches-valid="0" branch-rate="0" complexity="0" version="1.9" timestamp="1642329736">
    <sources>
        <source>/builds/fh1ch/hello-inline-coverage-rust</source>
    </sources>
    <packages>
        <package name="src/main.rs" line-rate="0.6875" branch-rate="0" complexity="0">
            <classes>
                <class name="main" filename="src/main.rs" line-rate="0.6875" branch-rate="0" complexity="0">
                    <methods>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0" branch-rate="0" complexity="0">
                            <lines>
                                <line number="22" hits="0">
                                </line>
                                <line number="23" hits="0">
                                </line>
                                <line number="24" hits="0">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::run" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0" branch-rate="0" complexity="0">
                            <lines>
                                <line number="22" hits="0">
                                </line>
                                <line number="23" hits="0">
                                </line>
                                <line number="24" hits="0">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main::test_run" signature="" line-rate="1" branch-rate="0" complexity="0">
                            <lines>
                                <line number="17" hits="1">
                                </line>
                                <line number="18" hits="1">
                                </line>
                                <line number="19" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::run" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main::test_run::{closure#0}" signature="" line-rate="1" branch-rate="0" complexity="0">
                            <lines>
                                <line number="16" hits="1">
                                </line>
                            </lines>
                        </method>
                    </methods>
                    <lines>
                    </lines>
                </class>
            </classes>
        </package>
    </packages>
</coverage>
```

**After:**

```xml
<?xml version="1.0"?>
<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>
<coverage lines-covered="11" lines-valid="16" line-rate="0.6875" branches-covered="0" branches-valid="0" branch-rate="0" complexity="0" version="1.9" timestamp="1642950173">
    <sources>
        <source>/builds/fh1ch/hello-inline-coverage-rust</source>
    </sources>
    <packages>
        <package name="src/main.rs" line-rate="0.6875" branch-rate="0" complexity="0">
            <classes>
                <class name="main" filename="src/main.rs" line-rate="0.6875" branch-rate="0" complexity="0">
                    <methods>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0" branch-rate="0" complexity="0">
                            <lines>
                                <line number="22" hits="0">
                                </line>
                                <line number="23" hits="0">
                                </line>
                                <line number="24" hits="0">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main::test_run" signature="" line-rate="1" branch-rate="0" complexity="0">
                            <lines>
                                <line number="17" hits="1">
                                </line>
                                <line number="18" hits="1">
                                </line>
                                <line number="19" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main" signature="" line-rate="0" branch-rate="0" complexity="0">
                            <lines>
                                <line number="22" hits="0">
                                </line>
                                <line number="23" hits="0">
                                </line>
                                <line number="24" hits="0">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::run" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::main::test_run::{closure#0}" signature="" line-rate="1" branch-rate="0" complexity="0">
                            <lines>
                                <line number="16" hits="1">
                                </line>
                            </lines>
                        </method>
                        <method name="hello_inline_coverage_rust::run" signature="" line-rate="0.7777777777777778" branch-rate="0" complexity="0">
                            <lines>
                                <line number="1" hits="1">
                                </line>
                                <line number="2" hits="1">
                                </line>
                                <line number="3" hits="1">
                                </line>
                                <line number="4" hits="1">
                                </line>
                                <line number="5" hits="0">
                                </line>
                                <line number="7" hits="1">
                                </line>
                                <line number="8" hits="0">
                                </line>
                                <line number="9" hits="1">
                                </line>
                                <line number="10" hits="1">
                                </line>
                            </lines>
                        </method>
                    </methods>
                    <lines>
                        <line number="1" hits="1">
                        </line>
                        <line number="2" hits="1">
                        </line>
                        <line number="3" hits="1">
                        </line>
                        <line number="4" hits="1">
                        </line>
                        <line number="5" hits="0">
                        </line>
                        <line number="7" hits="1">
                        </line>
                        <line number="8" hits="0">
                        </line>
                        <line number="9" hits="1">
                        </line>
                        <line number="10" hits="1">
                        </line>
                        <line number="16" hits="1">
                        </line>
                        <line number="17" hits="1">
                        </line>
                        <line number="18" hits="1">
                        </line>
                        <line number="19" hits="1">
                        </line>
                        <line number="22" hits="0">
                        </line>
                        <line number="23" hits="0">
                        </line>
                        <line number="24" hits="0">
                        </line>
                    </lines>
                </class>
            </classes>
        </package>
    </packages>
</coverage>
```

</details>

The only real difference besides some method order rearrangement is that the class level lines (`coverage > packages > package > classes > class > lines`) are now populated, containing the full line coverage data of the tested file. 

As mentioned already, while the cobertura format has a strictly defined structure (see [the official dtd definition](https://github.com/cobertura/cobertura/blob/master/cobertura/src/test/resources/dtds/coverage-04.dtd)), the actual semantic of each entry is nowhere to be found in written form. So the following part is not based on hard facts, but rather based on observations made when using different languages and frameworks and implementing [GitLab's code coverage feature](https://about.gitlab.com/releases/2020/09/22/gitlab-13-4-released/#inline-code-coverage-remarks-inside-mr-diffs):

Cobertura doesn't know the concept of *orphaned line* but rather reports the line coverage per *method* as well as *per class*. This can be seen when looking at reports created by Coberura itself (e.g. the [Jenkins Cobertura plugin example](https://github.com/jenkinsci/cobertura-plugin/blob/master/src/test/resources/hudson/plugins/cobertura/coverage-with-data.xml#L39); see the identical line information on method and on class level) as well looking into the [Coberura XML reporter class](https://github.com/cobertura/cobertura/blob/master/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java#L250), which dumps the complete line coverage information on class level. Exactly this part is used by GitLab to analyse the per-file line coverage and if missing, no coverage is reported at all.

Looking at that class level line coverage, the `gcovr` cobertura report actually seems to work as expected and creates proper hit counts when added to GitLab. Of course having test coverage for all of this should be the gaol, but we first have to define what's actually expected as output.

svenfoo:
Thanks for the detailed explanation of the change and thanks for the linked examples. The change looks good to me and I'd approve it.

marco-c (approved):
Thanks everyone, especially PR author ;)

svenfoo:
Would be awesome to have a binary release for 0.8.6, as that makes it much easier to use grcov from gitlab pipelines.

snue:
> Would be awesome to have a binary release for 0.8.6, as that makes it much easier to use grcov from gitlab pipelines.

0.8.6 is available with the fix from crates.io already. (Thanks for the quick merge/release!) We are picking it up via regular `cargo install grcov` in our GitLab pipeline.

svenfoo:
Yes, thanks, I know it can be done that way. Downloading the pre-compiled binary is faster though and the `coverage` pipeline is already so slow.

snue:
Ah got it. Nevermind then.

marco-c:
It looks like there was a failure in the job that uploads the binaries, #766.

marco-c:
The release is now available: https://github.com/mozilla/grcov/releases/tag/v0.8.6.

marco-c:
> I can also provide a follow-up MR, which adds a [GitLab coberatura example to the readme](https://github.com/mozilla/grcov/issues/468#issuecomment-823544526), if this MR goes into the right direction.

@fh1ch would you like to do that?

fh1ch:
> @fh1ch would you like to do that?

@marco-c sure thing! I was already working on it, just had to get my verification pipeline to get green.

I would propose a single GitLab example, showcasing the overall coverage recording (sole percentage value) as well as the inline coverage reporting (MR coverage per line), all source coverage based. Is using the pre-built `gcovr` ok for you (see [example](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/merge_requests/5/diffs))? It seems that installing `gcovr` using cargo in a rust:nightly image currently fails (see [example](https://gitlab.com/fh1ch/hello-inline-coverage-rust/-/jobs/2045888856)). Alternatively, I could also use a stable rust version, this seems to run just fine.

If there is no objection from your end, I'll provide an MR over the next few days for that.

snue:
> It seems that installing gcovr using cargo in a rust:nightly image currently fails

I have seen this error before. It fails because this compilation is already running with `RUSTFLAGS=-Zinstrument-coverage`. In my experience it will compile fine if you set the `RUSTFLAGS` only before building/testing the actual project. That means you cannot set it in the `.gitlab-ci.yml` variables section anymore of course.

But this might give issues for collecting coverage for `grcov` itself.

Also from the `grcov` Readme:
> Alternatively, you can export RUSTC_BOOTSTRAP=1, which basically turns your stable rustc into a Nightly one.

Maybe that's a preferable choice for the example? (But looking at your `.gitlab-ci.yml` you already do that :+1: )

PR #781, opened 2022-02-22
Don't die on invalid utf-8 in source code

russelltg:
This allows report generate when source code is not valid utf-8, just replacing invalid characters with the unicode replacement character.


marco-c (dismissed):
LGTM, @calixteman what do you think?

src/html.rs
```diff
@@ -364,7 +364,7 @@ fn gen_html(
                 .map(|&v| v as i64)
                 .unwrap_or(-1);
 
-            (index, count, l.unwrap())
+            (index, count, String::from_utf8_lossy(&l.unwrap()).into_owned())
```
calixteman, line 367: If I understand correctly, `from_utf8_lossy` is applied to each line of text. Why not apply it one time to all the text ?
Generally speaking, we should probably either detect the encoding or add an option to specify it and then use the correct decoder.

src/html.rs
```diff
@@ -354,7 +354,7 @@ fn gen_html(
     ctx.insert("branch_enabled", &branch_enabled);
 
     let items = f
-        .lines()
+        .split(b'\n')
```
calixteman, line 357: Why this change ?
It isn't the same as `lines`:
https://doc.rust-lang.org/nightly/std/io/trait.BufRead.html#method.lines

calixteman: We should have some tests to cover different kind of eols.

src/html.rs
calixteman: About detecting encoding, I wrote this code few years ago:
https://github.com/mozilla/rust-parsepatch/blob/b3dbb48ecc73de3fecea20c7e8848a3fc2239772/tests/test.rs#L188-L201

russelltg: I kept it on a per-line basis to avoid allocating a buffer for the whole file and keep the streaming nature of it, but certainly could change it to happen per-file.

Encoding detection wise, I could add a similar sort of detection, it was not necessary for my case, so I didn't think about it.

src/html.rs
russelltg: It's not quite the same as lines, in that lines just returns an error if it's invalid UTF-8, versus split gives you `[u8]`'s that you can call `from_utf8_lossy` on. 

I have missed that `Lines` discards trailing CR characters at the end of lines (https://doc.rust-lang.org/nightly/src/std/io/mod.rs.html#2791-2810) so if we stick with `split` that'll have to be fixed.

calixteman:
I think we can live with utf-8 only for now and if some people complain we could then try to detect the encoding.
But if something is wrong when decoding we should emit a warning (one time is enough: no need to overflow the stdout).

About line decoding vs file decoding: I've no strong opinion on that, so you can keep it like you did.
And the line breaking must be fixed: it should enough to split on "\n" and then right-strip a potential "\r".

src/html.rs
```diff
@@ -407,7 +422,7 @@ fn gen_html(
                 .map(|&v| v as i64)
                 .unwrap_or(-1);
 
-            (index, count, l.unwrap())
+            (index, count, l.to_string())
```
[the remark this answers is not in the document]
russelltg: Because we are now using `String::lines`, which yields `&str`'s, instead of `BufReader::lines` which yields `Result<String>`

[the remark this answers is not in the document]
calixteman: Oh yes, you're right.
Do we have to use `to_string` here ? or maybe we can just use `l`.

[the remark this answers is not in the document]
russelltg: Great catch. `l` is just fine--didn't realize that `ctx` was just json-serializing it

calixteman (approved):
LGTM. Thank you very much.

PR #782, opened 2022-02-24
Fix coverage percentage calculation for badges

### LINKED ISSUE #763: 0.8.5 always reports 0% in coverage.json
It seems the coverage.json file generated for shields.io is broken in 0.8.5, as it now always reports 0% coverage. The HTML file created is still correct and shows the right coverage stats. Reverting to 0.8.4 fixes it, and the project then shows the correct percentage again.

With 0.8.5:
```json
{"schemaVersion":1,"label":"coverage","message":"0%","color":"red"}
```

And 0.8.4:
```json
{"schemaVersion":1,"label":"coverage","message":"24%","color":"red"}
```

Args:
```
grcov . --binary-path ./target/debug/ -s . -t html --branch --ignore "*cargo*" -o ./coverage.xml --excl-br-start "mod tests \{" --excl-start "mod test \{"  --prefix-dir /src/ .
```

Project is in Rust, with version `rustc 1.60.0-nightly (84322efad 2022-01-23)`, with latest llvm-tools-preview. It happens with both profraw files and gcda/gcno files.

feroldi:
This patch fixes an issue with the calculation of coverage percentage for badges. The
issue is that all badges were being generated with a `0%` percentage on
them, no matter the coverage. This was due to an integer division that
always resulted in zero. Doing a floating-point division instead solves
it.

Closes #763

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -5,7 +5,6 @@ use std::collections::HashMap;
 use std::collections::{btree_map, BTreeMap};
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader, Write};
-use std::ops::{Div, Mul};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use tera::try_get_value;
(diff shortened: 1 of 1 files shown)


feroldi:
I have no idea what these failed tests mean. I haven't touched anything that could break so many of them.

marco-c:
I think the test failures are unrelated to your patch.

marco-c (approved):

marco-c:
Thanks!

Issue #631, opened 2021-06-02
Feature request: Add configurable thresholds for the HTML output


rianquinn:
If there some way to make the following configurable?

```rust
    let conf = Config {
        hi_limit: 90.,
        med_limit: 75.,
        fn_hi_limit: 90.,
        fn_med_limit: 75.,
        branch_hi_limit: 90.,
        branch_med_limit: 75.,
        date: Utc::now(),
    };
```

The project that we use this for needs 100% for lines and doesn't care about the others.

calixteman:
We could add an option `--config-file` for example which could take a json file (there's already a dep on serde-json) containing those options and some others we've in `--help`.
Would you mind making a PR for such thing ?

rianquinn:
I have never programmed in Rust before, I am sure that I can manage, so yes, I will take a stab at it. The other feature that would be really nice is some way to see the branch coverage information. This is something that is needed for critical systems compliance, and there really isn't any open source tools that handle this properly. 

Specifically, with C++. Suppose you have the following:
```cpp
template<typename T>
bool foo(int val)
{
    if (val != 42) {
        return false;
    }

    return true;
}
```

This simple function appears to only have two branches (take or do not take the if), when in fact, GCOV will report two branches for every instantiated version of `foo`. For for example, `foo<int>()` is 2, and `foo<bool>()` would be another 2. GRCOV right now just shows the branch % with no real information about why the branches are lower than they should be. 

One feature that would really set GRCOV apart from all of the other tools would be useful information around branches. This would include:
- With all of the template versions merged, are all of the branch combinations hit? This is useful when you are using template to mock inputs. You don't care about testing every version of a branch, you just want to make sure that every branch is covered across all of your unit tests. 
- Information about which branches are not hit (given all possible combinations) and which specific branch was not taken. For example, LCOV and GCOVR will tell you branch 0, 1 and 3 were taken, but branch 2, 4 and 5 were not. The only way to figure out what this means is to look at the raw GCOV output and demangle the C++ functions names and then manually figure out which template instantiation belongs with each branch. Its not scalable at all. It would be great if GRCOV had the same branch taken, not taken, not executed data that LCOV and GCOVR have, but when you hover over one of the branch results, it highlights the specific portion of the branch that it belongs to, and then also shows a popup with the demangled C++ name of the function so that you can see specifically which version of template that branch is referring too. Without this information, branch data is simply impossible to use with C++. 

[...]

PR #783, opened 2022-02-27
Add output config file param for html outptut (#631)

chrta:
The intend is that the new command line parameter `--output-config-file` could also be used by other exporters later, so it has a pretty generic name.

With this is is now possible to configure the html exporter.
The default thresholds can be overwritten. Also different tera templates can be used.

Example config file:
```
{
    "hi_limit": 93,
    "med_limit": 75,
    "fn_med_limit": 75,
    "templates": {
	"macros.html": "src/templates/macros.html",
	"base.html": "src/templates/base.html"
    }
}
```


marco-c:
@calixteman @lu-zero any opinion here?

lu-zero:
The idea is fine to me. Tomorrow I'll look at the code in detail.

lu-zero (approved):
It looks fine to me also code-wise.

Issue #570, opened 2021-01-28
floating point type error with gcov v9.3.0 


wideglide:
I'm geting a floating point error with grcov v0.7.1 and gcc/gcov version 9.3.0 on Ubuntu 20.04.

```shell
$ /usr/local/bin/grcov --branch -o outputs/cov/gcov/coverage.info gcov
22:38:25 [ERROR] A panic occurred at src/parser.rs:371: called `Result::unwrap()` on an `Err` value: Error("invalid type: floating point `7809720`, expected u64", line: 1, column: 84)
22:38:25 [ERROR] A panic occurred at src/parser.rs:371: called `Result::unwrap()` on an `Err` value: Error("invalid type: floating point `1864090`, expected u64", line: 1, column: 24437)
```

marco-c:
What is `gcov` in your cmd line? Is it the `gcov` executable from GCC?
You shouldn't pass it to grcov, you should instead pass the paths to your coverage artifacts.

wideglide:
`gcov` is a directory with the source files and coverage artifacts.

marco-c:
@wideglide can you try running grcov on one artifact at a time? This way you can identify which one is breaking and upload it here.

Can you try with GCC 8 or GCC 10, to confirm it's restricted to GCC 9?

PR #803, opened 2022-03-30
Handle floats gracefully when expecting u64 in parser.rs

oskirby:
We have been experiencing difficulty using grcov in our functional tests for the [mozilla-vpn-client](https://github.com/mozilla-mobile/mozilla-vpn-client). We have been finding that when a test runs for a sufficiently long time, the counters start to get large and the JSON deserializer is mistaking them for `f64` types instead, which eventually leads to panics.

As a workaround for this, I propose the following patch, which attempts to convert these unexpected floats back into a `u64` during deserialization. Many thanks to @Nils-TUD for his investigation into the issue.

Fixes: #570


marco-c:
LGTM, could you add a test case? Using a real file from the mozilla-vpn-client tests would be nice.

marco-c:
To test, we should actually call the parse_gcov_gz function on the output of gcov with those gcno/gcda files.
So:
1) Run `gcov -i PATH_TO_GCNO`
2) Get the output and store it in a file which you add to this repo
3) Add a test that loads the file from 2 and runs parse_gcov_gz on it

marco-c:
I got the relevant file from @oskirby, I'll finish the test and then merge this.

marco-c:
Here's an extract from the JSON:
```
{"branches": [], "count": 39, "line_number": 141, "unexecuted_block": false, "function_name": "_ZNK17QArrayDataPointerI10ServerCityE4dataEv"}, {"branches": [], "count": 205, "line_number": 141, "unexecuted_block": false, "function_name": "_ZNK17QArrayDataPointerI13ServerCountryE4dataEv"}, {"branches": [], "count": 1.37297e+06, "line_number": 141, "unexecuted_block": false, "function_name": "_ZNK17QArrayDataPointerIDsE4dataEv"}, {"branches": [], "count": 41341, "line_number": 141, "unexecuted_block": false, "function_name": "_ZNK17QArrayDataPointerIcE4dataEv"},
```

Clearly a GCC bug, multiple values for the same line.

marco-c (approved):

PR #818, opened 2022-04-25
Added missing "%" to branches.

Nils-TUD:


calixteman (approved):
Thank you for fixing that.

PR #840, opened 2022-06-07
WIP: Make the HTML report more accessible

federicomenaquintero:
We are using grcov for the [accessibility infrastructure modules in GNOME](https://gitlab.gnome.org/GNOME/at-spi2-core).  I'm starting this PR to make the generated HTML report easier to navigate for people who use screen readers.

Grcov's per-file reports indicate line coverage by color-coding and showing the number of times each line got executed, so they are easy to scan quickly by sighted users.  Each line has the line number, the coverage info, and the source code; the HTML for each line is a `<div>` with other `<div>`s for each column.

For users of screen readers, I've started with a few changes for navigation:

* Wrap all the lines in a `<div role="table">`.  This makes screen readers aware that this has tabular structure, and makes the line report easy to navigate to.  It also gives users a hint of how to navigate within the report.
* Give each line's `<div>` a `role="row"`, since it is now part of a table.
* Give each column a `role="cell"`.

With the above, each logical cell can be navigated using a screen reader's usual hotkeys for tables.

Next, to allow fast scanning, give the coverage column an `aria-label` with values as follows:

* `"no coverage"` - lines that are not executable.  Is this too long a string?  Screen readers allow interruption if you move to the next row before it finishes saying "no coverage", but I'm open to suggestions.
* `"0"` - lines that did not get executed - the ones to watch for, presumably.
* `"some-number"` - the actual number of times the line got executed.

One thing I'm not sure about - should the aria-label also have an indication of what the number means?  I.e. `aria-label="0 times"` instead of `aria-label="0"`?  The whole table does not have column headings; would that be better / less verbose instead?

I have not tweaked the `index.html` summary template.  That one is already a `<table>`, thankfully, but I wonder if the labels could be better: `42 out of 45 lines covered" instead of `42 / 45` for screen readers, for example.  I'm not sure.

Maybe the progress bars in the second column should be hidden for screen readers?  They only duplicate the numerical information from the third column.

### DIFF
--- a/src/templates/file.html
+++ b/src/templates/file.html
@@ -5,6 +5,7 @@
 
 {% block content -%}
     {{ macros::summary(parents=parents, stats=stats) }}
+    <div role="table" aria-label="Coverage report">
     {%- for item in items -%}
         {%- if item.1 > 0 -%}
             {%- set highlight = "success" -%}
@@ -34,5 +35,6 @@
             </div>
         </div>
     {%- endfor -%}
+    </div>
 {% endblock content -%}
 
--- a/src/templates/file.html
+++ b/src/templates/file.html
@@ -11,14 +11,17 @@
             {%- set highlight = "success" -%}
             {%- set highlight_light = "success-light" -%}
             {%- set count = item.1 -%}
+            {%- set aria_label = count -%}
         {%- elif item.1 < 0 -%}
             {% set highlight = "white" -%}
             {% set highlight_light = "white" -%}
             {% set count = "" -%}
+            {%- set aria_label = "no coverage" -%}
         {%- else -%}
             {%- set highlight = "danger" -%}
             {%- set highlight_light = "danger-light" -%}
             {%- set count = "" -%}
+            {%- set aria_label = "0" -%}
         {%- endif -%}
         <div class="columns p-0 m-0" role="row">
             <div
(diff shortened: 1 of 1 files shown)


lu-zero:
It seems to me good already, if you can attach a rendered html with your changes is easier to review it thought.
If adding aria labels with longer explanations per-cell helps the experience go for it.

lu-zero (approved):
It seems good to me.

marco-c:
Thanks for the PR!

> One thing I'm not sure about - should the aria-label also have an indication of what the number means? I.e. `aria-label="0 times"` instead of `aria-label="0"`? The whole table does not have column headings; would that be better / less verbose instead?

I would keep just "0" to be less verbose.

> I have not tweaked the `index.html` summary template. That one is already a `<table>`, thankfully, but I wonder if the labels could be better: `42 out of 45 lines covered" instead of `42 / 45` for screen readers, for example. I'm not sure.

Probably "42 / 45" is good enough given that the user knows the context, but being more explicit can't hurt. I'll leave the decision to you :)

> Maybe the progress bars in the second column should be hidden for screen readers? They only duplicate the numerical information from the third column.

Yeah, I'd say it doesn't make much sense to keep them visible to screen readers.

marco-c (approved):
I'll approve it for now, but I won't merge it yet to allow you to update it with the other changes you were planning.
The proposed changes are already a net improvement though, so we can also merge this and work on follow-ups later.

federicomenaquintero:
Thanks for the merge!  I think we'll know over time whether the reports are more usable for screen reader users.  I'll submit other PRs if stuff comes up.

PR #882, opened 2022-09-10
Update dependencies and remove useless fomat-macros

calixteman:


marco-c (approved):

marco-c:
fomat-macros was added in #297.

Issue #498, opened 2020-10-23
Add a Text/Markdown output


lu-zero:
It would be nice to have an output that's readable from the shell.

dbr:
I would have found something like this very useful when setting up coverage in a CI system - it's a helpful sanity check to have a short summary of coverage results in the log output instead of having to (fully) rely on the report (which with something liek codecov could, possibly, fail to upload)

[coverage.py's](https://coverage.readthedocs.io/en/coverage-5.5/) default report output is a good example:

```
$ coverage report -m
Name                      Stmts   Miss  Cover   Missing
-------------------------------------------------------
my_program.py                20      4    80%   33-35, 39
my_other_module.py           56      6    89%   17-23
-------------------------------------------------------
TOTAL                        76     10    87%
```

Personally I could happily remove the `Missing: 33-35, 39` output column

marco-c:
This would be just a matter of adding a new output format.

PR #885, opened 2022-09-15
Add markdown output

nicolas-geniteau:
This PR add a new markdown output. This is related to https://github.com/mozilla/grcov/issues/498.

Example output:

## RAW
```
| file          | coverage | covered | missed_lines |
|---------------|----------|---------|--------------|
| foo/bar/a.cpp | 100%     | 2 / 2   |              |
| foo/bar/b.cpp | 40%      | 2 / 5   | 1, 5-7       |

Total coverage: 57%
```

## Markdown parsed

| file          | coverage | covered | missed_lines |
|---------------|----------|---------|--------------|
| foo/bar/a.cpp | 100%     | 2 / 2   |              |
| foo/bar/b.cpp | 40%      | 2 / 5   | 1, 5-7       |

Total coverage: 57%

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -3,7 +3,7 @@ use md5::{Digest, Md5};
 use rustc_hash::FxHashMap;
 use serde_json::{self, json, Value};
 use std::cell::RefCell;
-use std::collections::{hash_map, BTreeSet};
+use std::collections::{hash_map, BTreeMap, BTreeSet};
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::{self, BufWriter, Read, Write};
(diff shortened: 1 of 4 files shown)


src/output.rs
```diff
@@ -856,4 +923,44 @@ mod tests {
...
+                },
+            ),
+        ];
+
+        let results = Box::new(results.into_iter());
+        output_markdown(results, Some(&file_path));
+
+        let results = &read_file(&file_path);
+        let expected_path = PathBuf::from("./test/").join(&file_name);
+        let expected = &read_file(&expected_path);
```
marco-c, line 962: Since it's just a few lines, could you put the expected output in a string in this test directly?

marco-c:
Can we output "covered" instead of "missed"?

src/output.rs
nicolas-geniteau: Updated

nicolas-geniteau:
@marco-c Thanks for your review. I just updated to output "covered" instead of "missed". Is it OK for you now ?

marco-c (approved):
LGTM, thanks!

PR #935, opened 2022-12-04
Replace infer with file-format

mmalecot:
Pros of `file-format` against `infer`:

- Support for many more formats
- Support for formats based on **MS-DOS Executable** more accurate.
- Enum-based API
- Zero-dependency


calixteman:
It'd be nice to have something explaining this PR...
What are the pros/cons for each of the crates: `infer` vs `file-format` ?

PR #967, opened 2023-02-25
Add the possibility to dump a gcno and an optional gcda to json

calixteman:
A gcno file contains some interesting information about the structure of the functions making the source file.
So this patch aims to provide a json dump of a gcno in order to make it consumable by other tools.


src/main.rs
```diff
@@ -318,6 +324,18 @@ fn main() {
         );
     }
 
+    if let Some(gcno) = opt.gcno {
```
marco-c, line 327: Should we make this option conflict with others, given that we return early when it is set?

marco-c (approved):

src/main.rs
calixteman: How do you want to do that ?

PR #978, opened 2023-03-08
Make html branch info dependent on cmdline argument --branch

edmundoferreira:
Currently if the command line option `--branch` is not passed, when generating html reports (`-t html`), the report still shows 100% all green in the _Branches_ html table column. Like on the screenshot below.

image

I believe that the html report should take the command line argument `--branch` into consideration, and not show any _Branches_ column in the report, if the variable is not passed.
So the current PR addresses this by modifying the `tera` templates to reuse the already existing `branch_enabled` variable inserted into the context. 

1. If the `--branch` is **not** passed, then the report will look like:
image

2. if the `--branch` is passed, then the report will look like (given that the tests were compiled with branching information i.e. GCC `-coverage` or older `-fprofile-arcs`):
image

### DIFF
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -11,7 +11,9 @@
                 <th>{{ kind }}</th>
                 <th class="has-text-centered" colspan="3">Line Coverage</th>
                 <th class="has-text-centered" colspan="2">Functions</th>
-                <th class="has-text-centered" colspan="2">Branches</th>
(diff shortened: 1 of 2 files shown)


edmundoferreira:
Moved to draft! Waiting merge of https://github.com/mozilla/grcov/pull/979

marco-c (approved):
Thanks!

PR #979, opened 2023-03-08
Fix clippy linting

edmundoferreira:
Fixes failed checks in many PRs for example https://github.com/mozilla/grcov/actions/runs/4345920963/jobs/7591298097

- fix all https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
- fix all https://rust-lang.github.io/rust-clippy/master/index.html#iter_kv_map
- allow https://rust-lang.github.io/rust-clippy/master/index.html#manual_filter since it's a false positive that is already fixed upstrean and will be  present on next clippy release

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -351,9 +351,7 @@ fn main() {
     let source_root = opt
         .source_dir
         .filter(|source_dir| source_dir != Path::new(""))
-        .map(|source_dir| {
-            canonicalize_path(&source_dir).expect("Source directory does not exist.")
-        });
+        .map(|source_dir| canonicalize_path(source_dir).expect("Source directory does not exist."));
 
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -474,6 +472,10 @@ fn main() {
     let output_path = match output_types.len() {
         0 => return,
         1 => opt.output_path.as_deref(),
+        //TODO: Remove false positive clippy warning when
```
marco-c, line 475: Could you file an issue so we don't forget to do this?

marco-c (dismissed):

src/main.rs
edmundoferreira: Yes I can make an issue and also added it the TODO for reference.
On it.

marco-c: #980

marco-c (approved):

PR #981, opened 2023-03-10
Add precision cmdline arg

edmundoferreira:
Currently multiple report types have different precision values. For example:
1. `markdown.md` has integer precision i.e. `12%`
2. `html/index.html` has 2 decimal precision i.e. `12.34%` but `html/coverage.json` only has integer precision i.e. `12%`
3. `covdir` has actual 2 decimal precision i.e. `12.23%` with some subtleties of the `serder_json`

The current PR:
1. create a command line arg `--precision` for the user to control the precision of all the reports types
2. propagates this precision to `markdown.md`
3. propagates this precision to `html/*` 
4. propagates this precision to `covdir`
5. no action on other output formats since they seem to have their own coverage precision formats.

### DIFF
--- a/src/covdir.rs
+++ b/src/covdir.rs
@@ -4,13 +4,13 @@ use std::collections::BTreeMap;
 pub use crate::defs::*;
 
 impl CDStats {
-    pub fn new(total: usize, covered: usize) -> Self {
+    pub fn new(total: usize, covered: usize, precision: usize) -> Self {
         let missed = total - covered;
         Self {
             total,
             covered,
             missed,
-            percent: Self::get_percent(covered, total),
+            percent: Self::get_percent(covered, total, precision),
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/covdir.rs
```diff
@@ -4,13 +4,13 @@ use std::collections::BTreeMap;
 pub use crate::defs::*;
 
 impl CDStats {
-    pub fn new(total: usize, covered: usize) -> Self {
+    pub fn new(total: usize, covered: usize, precision: usize) -> Self {
```
calixteman, line 7: Why do you want to use an usize for the precision ?

edmundoferreira: I debated this same question myself. 

Specially after digging a bit more on [the book](https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-types) (I'm have more experience with C++)
> The primary situation in which you’d use isize or usize is when indexing some sort of collection.

However the reason I used `usize` was that I first did the `markdown.rs` first and there `format!()` is used, like:
```
coverage: format!("{:.precision$}%", (covered as f32 * 100.0 / result.lines.len() as f32), ),
```
 Which according to the [docs](https://doc.rust-lang.org/std/fmt/#width)
> The value for the width can also be provided as a `usize` in the list of parameters by adding a postfix $, indicating that the second argument is a `usize` specifying the width.

Then the `usize` type propagated to the other files (`html.rs`, `covdir.rs`,...)

Again I was about to rebase on another type like `u8` or `u32`, but didn't to keep consistency with `std:format` `width`. 

However I'm totally fine rebasing and changing the type everywhere.
Any feedback is welcome. To be honest the only fundamental thing that I wanted to achieve is consistent formatting across the report outputs. If the `mozilla/grcov` maintainers prefer to have this setting in a variable and not exposed as another cmd line argument to the users, I'm also fine with that.

PR #982, opened 2023-03-11
Use create_dir_all instead of create_dir for html output

### LINKED ISSUE #972: Adding sub-path to output-path option does not create the directory path
Hello, when I run the following command and the directory `coverage` does not exist:
`grcov . --binary-path ./target/debug/deps -s . -t html --branch --ignore-not-existing --ignore ../* --ignore /* --ignore tools/xtask/* --ignore */src/tests/* -o coverage/subpath`
The tool does not give any html output and neither warns the user about the missing directory (I also tried with the option `--log-level INFO`
I think the tool should either create the directory path or either panic.

**Environment**
Windows 11
rustc version: 1.64.0
grcov version: 0.8.13

GioPat:
Fixes #972.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -530,7 +530,7 @@ pub fn output_html(
             eprintln!("{} is not a directory", output.to_str().unwrap());
             return;
         }
-    } else if std::fs::create_dir(&output).is_err() {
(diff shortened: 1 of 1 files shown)


marco-c (approved):
Thanks!

PR #996, opened 2023-03-22
Add Gitlab CI example

### LINKED ISSUE #814: Add a GitLab cobertura example to the README

willmurnane:
Closes #814.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ build:
   script:
     # Run all your Rust-based tests
     - cargo test --workspace
-    # Optionally, run some other command that exercises your code:
+    # Optionally, run some other command that exercises your code to get more coverage:
     - ./bin/integration-tests --foo bar
     # Once a release with https://github.com/mozilla/grcov/pull/893 is available, modify this to use the multiple-outputs mechanism :)
     - grcov target/coverage --binary-path target/debug -s . -o target/coverage.xml -t cobertura --keep-only 'src/*'
@@ -337,7 +337,7 @@ This also ties into Gitlab's coverage percentage collection, so in merge request
 - increases or decreases of coverage
 - whether particular lines of code modified by a merge request are covered or not.
 
-Additionally, the HTML-formatted coverage report (if you leave it enabled) will be produced as an artifact. 
+Additionally, the HTML-formatted coverage report (if you leave it enabled) will be produced as an artifact.
 
 ### Alternative reports
 
--- a/README.md
+++ b/README.md
@@ -304,22 +304,24 @@ script:
 
 #### grcov with Gitlab
 
-Here is an example `.gitlab-ci.yml` which will build your project, then collect coverage data in a format that Gitlab understands:
+Here is an example `.gitlab-ci.yml` which will build your project, then collect coverage data in a format that Gitlab understands. It is assumed that you'll use an image which already has relevant tools installed, if that's not the case put the appropriate commands at the beginning of the `script` stanza.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -301,6 +302,31 @@ script:
...
 ```
 
+#### grcov with Gitlab
+
+Here is an example `.gitlab-ci.yml` which will build and collect coverage data in Gitlab:
+
+```yaml
+build:
+  script:
+    - cargo test --workspace
```
marco-c, line 312: Can you add LLVM_PROFILE_FILE here as well?

README.md
```diff
@@ -301,6 +302,31 @@ script:
...
+    # Optionally, run some other command that exercises your code:
+    - LLVM_PROFILE_FILE="target/coverage/%p-%m.profraw" ./bin/integration-tests --foo bar
+    # Once a release with https://github.com/mozilla/grcov/pull/893 is available, modify this to use the multiple-outputs mechanism :)
+    - grcov target/coverage --binary-path target/debug -s . -o target/coverage.xml -t cobertura
+    - grcov target/coverage --binary-path target/debug -s . -o target/coverage -t html
+    # This is an awful hack whose goal is to print the coverage number by extracting it from the HTML report.
+    - awk 'c&&--c { gsub(" *<abbr[^>]+>", "Coverage: ", $0); print; } /<p class="heading">Lines</p>/{c=2}' target/coverage/index.html
```
marco-c, line 319: Might be easier to extract from the markdown output?

willmurnane: Yes, it might be. `grcov` takes about 45 seconds per run on my project, though, so it's a noticeable slowdown to have to do even 2 runs of it. Thus, I used the HTML report to get the coverage rather than running another time. Eagerly awaiting a release with multi-output mode :)

The xml output is also probably easier to get the overall coverage from, but in my testing it looks like you'd want `--keep-only 'src/*'` or similar on that invocation to avoid getting a bunch of un-covered external libraries included in your coverage report. I can switch to that if you like, this MR was just assembled from what I'm actually doing, not necessarily thinking about the best thing to do.

I'll read over the changes with "what's the best way to do this" in mind.

README.md
willmurnane: Done, setting the `variables:` adds an environment variable so it'll be present for all the code execution.

marco-c (approved):
Thanks!

README.md
```diff
@@ -301,6 +302,43 @@ script:
...
+build:
+  variables:
+    # Set an environment variable which causes LLVM to write coverage data to the specified location. This is arbitrary, but the path passed to grcov (the first argument) must contain these files or the coverage data won't be noticed.
+    LLVM_PROFILE_FILE: "target/coverage/%p-%m.profraw"
+  script:
+    # Run all your Rust-based tests
+    - cargo test --workspace
+    # Optionally, run some other command that exercises your code to get more coverage:
+    - ./bin/integration-tests --foo bar
+    # Once a release with https://github.com/mozilla/grcov/pull/893 is available, modify this to use the multiple-outputs mechanism :)
```
marco-c, line 319: 0.8.14 is now available, care to update this? :)

willmurnane: It's not on `crates.io` yet, and I prefer not to point directly to github. Thank you for cutting a release, though, I'm looking forward to cutting some time off my obscene build times ;)

marco-c: 0.8.15 is now available on crates.io

PR #997, opened 2023-03-24
Switch `structopt` for `clap` v3

CosmicHorrorDev:
This switches things back to `clap` now that the derive changes landed

The changes for `LevelFilter` are a little awkward due to the orphan rule. I couldn't find a good way to have `clap` parser to a `LevelFilter` while enumerating possible values. It's possible that I just missed something

I opted for going to switch to v3 instead of jumping all the way to v4 to minimize controversial changes since v4 involved a lot of changes to help messages that would likely leave more to debate. The (imo confusingly named) `deprecated` feature enables deprecation messages, so the transition to v4 should be pretty smooth anyways

Lastly I couldn't get the full test suite to pass on master to begin with (couldn't find any documentation for setup I need either). At the very least all of the tests that pass on master passed on this PR for me and all the tests that failed seemed to fail in the same way

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@
 #[global_allocator]
 static GLOBAL: tcmalloc::TCMalloc = tcmalloc::TCMalloc;
 
+use clap::{ArgGroup, Parser};
 use crossbeam_channel::bounded;
 use log::error;
 use regex::Regex;
@@ -15,7 +16,6 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 use std::{process, thread};
-use structopt::{clap::ArgGroup, StructOpt};
 
 use grcov::*;
 
@@ -72,6 +72,7 @@ impl OutputType {
     }
 }
 
+#[derive(clap::ValueEnum, Clone)]
 enum Filter {
     Covered,
     Uncovered,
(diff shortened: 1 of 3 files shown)


lu-zero:
did you forget to initialize the submodules?

src/main.rs
```diff
@@ -89,23 +90,59 @@ impl FromStr for Filter {
     }
 }
 
-#[derive(StructOpt)]
-#[structopt(
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+struct LevelFilterArg(LevelFilter);
+
+impl clap::ValueEnum for LevelFilterArg {
```
lu-zero, line 96: why not using its `#derive` ?

lu-zero: Ah yes, it is in the wrong place, probably I could update https://github.com/lu-zero/arg_enum_proc_macro

lu-zero (approved):

src/main.rs
CosmicHorrorDev: The derive wouldn't work because it's a newtype wrapper. Not sure if `arg_enum_proc_macro` could support that without having to pass the all the inner type's variants in some way

marco-c (approved):
Thanks @CosmicHorrorDev for the contribution and thanks @lu-zero for the review ;)

marco-c:
@CosmicHorrorDev it'd be great if you could also submit a PR for upgrading to clap v4.

CosmicHorrorDev:
Sure! I'll work on that tomorrow :)

PR #1000, opened 2023-03-26
Update clap to v4

CosmicHorrorDev:
This is a followup to #997 

From an user's perspective this should largely mimic clap v3. The main differences now are that the help-text primarily uses non-colored styling now (I believe making this configurable is one of the goals of clap v5), and the flags and args are now sorted in definition order instead of alphabetically. Here's the start of the old and new short-help for comparison

image

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,7 @@ dependencies = [
  "once_cell",
  "strsim 0.10.0",
  "termcolor",
+ "terminal_size",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


lu-zero:
The tracking issue about colors should be https://github.com/clap-rs/clap/issues/3234

lu-zero (approved):
It looks fine to me.

marco-c:
As a follow-up, we should be moving the most frequently used arguments to the top.

marco-c:
As a follow-up, we should be moving the most frequently used arguments to the top.

PR #1004, opened 2023-04-02
Consolidate `--output-type` and `--output-types`

### LINKED ISSUE #1002: Option --output-path doesn't work after v0.8.15
In version 0.8.13 the following command produced `lcov.info` for me:
```bash
grcov target/coverage --binary-path target/debug/ --source-dir . -o lcov.info
```
But after update to 0.8.15 the command doesn't create it and exits with code 0. I believe it caused by migration from `structopt` to `clap`.

CosmicHorrorDev:
Fixes #1002

As discussed in #1002 this combines `--output-type` and `--output-types` using an alias and changes the default output type to lcov

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -153,6 +153,7 @@ struct Opt {
     llvm_path: Option<PathBuf>,
     /// Sets a custom output type.
     #[arg(
+        short = 't',
         long,
         long_help = "\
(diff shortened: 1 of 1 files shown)


lu-zero (approved):

marco-c (approved):
Thanks, this makes perfect sense.

Issue #1013, opened 2023-04-11
grcov sometimes takes very long to execute


jonhoo:
This is extracted from the discussion [in #326](https://github.com/mozilla/grcov/issues/326#issuecomment-1452770109).

In some projects, some of the time, when I execute `grcov`, it takes a very long time to execute, using 100% of one core for several minutes at the time. I executed it like this:
```
$ pgrep -a grcov
61819 grcov $X/cb-coverage-target/profraw --source-dir . --binary-path $X/cb-coverage-target/debug/ --branch --ignore-not-existing <a few --excl-* args> --keep-only src/* --keep-only src/* --keep-only src/bin/* -t html -o $X/coverage/
$ ls -la $X/cb-coverage-target/profraw
.rw-r--r-- 3.1M jongje 11 Apr 16:45 61284-8806487336122761277_0.profraw
.rw-r--r-- 2.6M jongje 11 Apr 16:45 61289-8403376454973083594_0.profraw
.rw-r--r--  24k jongje 11 Apr 16:45 61296-16724270415138201609_0.profraw
.rw-r--r-- 5.5M jongje 11 Apr 16:45 61297-4578242808341374556_0.profraw
.rw-r--r--  224 jongje 11 Apr 16:45 61378-18005603889831348665_0.profraw
.rw-r--r-- 4.6M jongje 11 Apr 16:45 61379-12535087358149662424_0.profraw
.rw-r--r-- 5.4M jongje 11 Apr 16:45 61458-13306843052265737977_0.profraw
.rw-r--r-- 5.5M jongje 11 Apr 16:45 61573-2063728991251543943_0.profraw
.rw-r--r-- 5.4M jongje 11 Apr 16:45 61659-2807099239873412591_0.profraw
.rw-r--r-- 336k jongje 11 Apr 16:45 61742-2870874979089868175_0.profraw
.rw-r--r-- 312k jongje 11 Apr 16:45 61742-5935495699248178733_0.profraw
.rw-r--r-- 543k jongje 11 Apr 16:45 61742-6234157094333981707_0.profraw
.rw-r--r-- 206k jongje 11 Apr 16:45 61742-8110670833310045757_0.profraw
.rw-r--r-- 8.3k jongje 11 Apr 16:45 61742-8156326955437846749_0.profraw
.rw-r--r--  27k jongje 11 Apr 16:45 61742-13964213668791941633_0.profraw
[...]

PR #1014, opened 2023-04-11
Recommend pointing binary-path at /deps

jonhoo:
This results in a significant speed-up since we can avoid invoking `llvm-cov export` on all the object files in `target/debug/incremental`.

See https://github.com/mozilla/grcov/issues/1013.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ RUSTC_BOOTSTRAP=1`, which basically turns your stable rustc into a Nightly one.
 Generate a html coverage report like this:
 
 ```sh
-grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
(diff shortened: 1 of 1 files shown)


TriplEight:
cc https://github.com/mozilla/grcov/pull/1015#issuecomment-1507071173

marco-c:
Given that we risk missing something by doing this, I would not recommend it by default in all examples in the README, but just add a note saying that it can improve performance

jonhoo:
It doesn't sound like it actually misses coverage if the reported coverage for build scripts is 100% anyway though (at least that was what one report said). But happy to update this to be a note about it improving performance instead if you'd prefer?

marco-c:
From https://github.com/mozilla/grcov/pull/1015#issuecomment-1507071173 it looks like build.rs files were not covered?

sandesh3407 (approved):

PR #1025, opened 2023-04-29
Add flag_name for coveralls output

jbreams:
I'm using grcov to generate coveralls output, and it was missing the `flag_name` parameter used in coverall's "parallel" mode that's used to tell coveralls that it should carry over any missing results that were in previous runs to this run (see the ["job" object description](https://docs.coveralls.io/api-jobs-endpoint#json-object-job) and [parallel build webhook](https://docs.coveralls.io/api-parallel-build-webhook) for more info). This PR just adds flag name as an optional command line argument.


marco-c (approved):
Thanks!

PR #1034, opened 2023-05-06
Update error message to indicate if a directory doesn't exist to dump coverage file

saurasm:
Modified function `get_target_output_writable` in `src/output.rs`.
Updated error message indicates if a directory doesn't exist for dumping the coverage file.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -50,14 +50,14 @@ pub fn get_target_output_writable(output_file: Option<&Path>) -> Box<dyn Write> 
                 if let Some(parent_path) = parent {
                     if !parent_path.exists() {
                         panic!(
-                            "Cannot create {} to dump coverage data, as {} doesn't exist",
+                            "Cannot create file {} to dump coverage data, as directory {} doesn't exist",
                             output.display(),
                             parent_path.display()
                         )
                     }
                 }
                 panic!(
-                    "Cannot create the file {} to dump coverage data.",
+                    "Cannot create file {} to dump coverage data, as the directory doesn't exist",
                     output.display()
                 )
             }))
--- a/src/output.rs
+++ b/src/output.rs
@@ -57,7 +57,7 @@ pub fn get_target_output_writable(output_file: Option<&Path>) -> Box<dyn Write> 
                     }
                 }
                 panic!(
-                    "Cannot create file {} to dump coverage data.",
+                    "Cannot create the file {} to dump coverage data.",
                     output.display()
                 )
             }))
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -50,14 +50,14 @@ pub fn get_target_output_writable(output_file: Option<&Path>) -> Box<dyn Write>
                 if let Some(parent_path) = parent {
                     if !parent_path.exists() {
                         panic!(
-                            "Cannot create {} to dump coverage data, as {} doesn't exist",
+                            "Cannot create file {} to dump coverage data, as directory {} doesn't exist",
```
marco-c, line 53:
```suggestion
                            "Cannot create file {} to dump coverage data, as the parent directory {} doesn't exist",
```

marco-c: I wonder if we can't just create the directory in this case

src/output.rs
```diff
@@ -50,14 +50,14 @@ pub fn get_target_output_writable(output_file: Option<&Path>) -> Box<dyn Write>
...
                             output.display(),
                             parent_path.display()
                         )
                     }
                 }
                 panic!(
-                    "Cannot create the file {} to dump coverage data.",
+                    "Cannot create file {} to dump coverage data, as the directory doesn't exist",
```
marco-c, line 60: In this case the error must be due to something else, as we cover the directory not existing case above

src/output.rs
```diff
@@ -50,14 +50,14 @@ pub fn get_target_output_writable(output_file: Option<&Path>) -> Box<dyn Write>
...
                         panic!(
-                            "Cannot create {} to dump coverage data, as {} doesn't exist",
+                            "Cannot create file {} to dump coverage data, as the parent directory {} doesn't exist.",
                             output.display(),
                             parent_path.display()
                         )
                     }
                 }
                 panic!(
-                    "Cannot create the file {} to dump coverage data.",
```
marco-c, line 60: Could you revert the change on this line?

saurasm: I already reverted that change. Line 60 in PR matches with [line 60 in the master branch](https://github.com/mozilla/grcov/blob/725054c044a59138e4670f10cf221d9d310e226d/src/output.rs#L60).

marco-c: There is a diff here, it looks like you removed "the "

saurasm: Sorry about that, I've added "the".

marco-c (approved):
Thanks!

PR #1056, opened 2023-06-12
Use CARGO_TARGET_DIR for target directory

L3nn0x:
Hello, if the target directory is not `target` then grcov will fail to run. This is a patch to correct this behavior.

Let me know if there is anything I need to do to get this merged in.

Thanks!


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grcov"
-version = "0.8.18"
```
marco-c, line 3: Could you revert this change? I'll bump the version in a later patch

L3nn0x: done!

marco-c:
There is a linting error related to formatting, could you run `cargo fmt`?

marco-c (approved):
Thanks!

PR #1177, opened 2024-02-24
Update navigation path in README.md

niStee:
This commit extends the navigation path in the README.md file to include "User Rights Assignment". This provides more precise instructions for users following the guide.


marco-c:
Landed in #1186, thanks for the PR!

PR #1178, opened 2024-03-22
Fix .gcda parsing for C++ templates

culhatsker:
When C++ code has templates, `gcov` tool may report same line more than once, thus `grcov` need to allow counting the sum of those values, not just accepting the last.


culhatsker:
@marco-c may I ask you to look at these changes?

marco-c:
@culhatsker could you add a test for this?

PR #1180, opened 2024-04-08
Cobertura conciseness and summary option

### LINKED ISSUE #556: output coverage summary (gitlab integration)
It would be great to have an easy way to output on stdout the coverage summary in a way [easily parsable by gitlab](https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-parsing).

rshearman:
- Add --print-summary option Add a --print-summary option along the lines
   of what gcovr supports and outputting in a compatible format. The number
   of functions covered isn't output (it is with gcovr) as this is more
   complicated to calculate and it's unclear if it would be useful.

   Fixes #556

 - Make cobertura output more concise and add cobertura-pretty output option 
   Some parsers of cobertura coverage files places limits on the file size
   (e.g. GitLab - 10MB), so being as concise as possible is advantageous.
   For this reason, change the cobertura output to not add newlines and
   indent between XML elements, and add a new output type cobertura-pretty
   for cases where the newlines and indent is desired.

   Changing the format for the existing cobertura output type rather than 
   keeping it the same and introducing a new concise output type was opted
   for since cobertura is designed to be machine-readable and most of the
   time users would not need to inspect the file manually.

   This results in a 42% reduction in output file size for a repository of
   mine.

 - Make output of plain lines in cobertura reports more concise For plain
   lines (without branch information) then there's no need to write out
   opening and closing tags for the line element. It is more concise to
   output an empty element, and conciseness is important to avoid hitting
   processing limits from certain services that parse the output earlier
   than is necessary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,6 +29,7 @@ enum OutputType {
     Covdir,
     Html,
     Cobertura,
+    CoberturaPretty,
     Markdown,
 }
 
@@ -45,6 +46,7 @@ impl FromStr for OutputType {
             "covdir" => Self::Covdir,
             "html" => Self::Html,
             "cobertura" => Self::Cobertura,
+            "cobertura-pretty" => Self::CoberturaPretty,
             "markdown" => Self::Markdown,
             _ => return Err(format!("{} is not a supported output type", s)),
         })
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -292,6 +297,9 @@ struct Opt {
     /// No symbol demangling.
     #[arg(long)]
     no_demangle: bool,
+    /// Print a summary of the results
+    #[arg(long)]
+    print_summary: bool,
```
marco-c, line 302: What if we have "summary" as one of the possible output types instead of adding a --print-summary option?

marco-c:
Thanks @rshearman, I've merged the two commits related to Cobertura.

PR #1221, opened 2024-09-25
Drop cargo_binutils dependency as it hasn't been updated for a while

marco-c:


src/llvm_tools.rs
```diff
@@ -146,12 +147,37 @@ pub fn profraws_to_lcov(
...
+    Ok(path)
+}
+
 fn get_profdata_path() -> Result<PathBuf, String> {
     let path = if let Some(mut path) = LLVM_PATH.get().cloned() {
         path.push(format!("llvm-profdata{}", EXE_SUFFIX));
         path
     } else {
-        Tool::Profdata.path().map_err(|x| x.to_string())?
+        llvm_tool_path("profdata").map_err(|x| x.to_string())?
```
calixteman, line 180: I suppose we can remove `Tool::Profdata` (it's only used there).

src/llvm_tools.rs
```diff
@@ -166,7 +192,7 @@ fn get_cov_path() -> Result<PathBuf, String> {
         path.push(format!("llvm-cov{}", EXE_SUFFIX));
         path
     } else {
-        Tool::Cov.path().map_err(|x| x.to_string())?
+        llvm_tool_path("cov").map_err(|x| x.to_string())?
```
calixteman, line 195: Same as above but for `Tool::Cov`.

calixteman (approved):
LGTM. Thank  you.

PR #1253, opened 2024-12-02
Update README.md

hfyeh:
Two modifications in `.gitlab-ci-yml`
1. Fix the .xml path so that the report can be found by GitLab
2. Replace ':' by '-' since YAML keeps ':' as an keyword

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,15 +374,15 @@ build:
         --output-types html,cobertura
 
     # Extract just the top-level coverage number from the XML report.
-    - xmllint --xpath "concat('Coverage: ', 100 * string(//coverage/@line-rate), '%')" target/coverage/cobertura.xml
-  coverage: '/Coverage: \d+(?:\.\d+)?/'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -374,15 +374,15 @@ build:
         --output-types html,cobertura
 
     # Extract just the top-level coverage number from the XML report.
-    - xmllint --xpath "concat('Coverage: ', 100 * string(//coverage/@line-rate), '%')" target/coverage/cobertura.xml
-  coverage: '/Coverage: \d+(?:\.\d+)?/'
+    - xmllint --xpath "concat('Coverage - ', 100 * string(//coverage/@line-rate), '%')" target/coverage/cobertura.xml
+  coverage: '/Coverage - \d+(?:\.\d+)?/'
```
marco-c, line 378: Is this actually necessary given it is within a string?

MarkusTieger: Yes, it is.

PR #1280, opened 2025-01-27
Improve jacoco parsing and improve kotlin support

### LINKED ISSUE #1279: jacoco parsing reports from kotlin
Kotlin, unlike java, does _not_ require a single top level class in the file, which generally means that the jacoco report, when parsed by grcov will simply error out. Looking more at the details, the sourcefile should be the key in the mapping and the classes should be assumed to be in that file.

Note, this will be more apparent when my PR is published.

jasongilanfarr:
Fixes #1279 

The jacoco parser made some relatively safe assumptions about the jacoco report, except when jacoco is used to instrument kotlin. Kotlin doesn't have the same restrictions as java when it comes to top level classes, so the report parsing will just immediately error out.

This improves the parser by making it sourcefilename centric instead of class centric (falling back on the pattern of top level class + .java == sourcefilename if the sourcefilename is missing).

In addition, when aggregated across multiple projects, the report commonly doesn't have the full relative paths, so we need to be able to find _both_ java and kotlin files. There is also an assumption that a fully qualified class can only exist in one location and it would panic; however, this isn't _entirely_ uncommon, so don't panic, but instead, inform the user that its not fully supported.


src/parser.rs
```diff
@@ -2019,7 +2011,7 @@ TN:http_3a_2f_2fweb_2dplatform_2etest_3a8000_2freferrer_2dpolicy_2fgen_2fsrcdoc_
         let mut branches: BTreeMap<u32, Vec<bool>> = BTreeMap::new();
         branches.insert(3, vec![true, true]);
         let expected = vec![(
-            String::from("hello.java"),
+            String::from("/hello.java"),
```
marco-c, line 2014: Wouldn't it be better to keep the trimming as it was to get `hello.java` here and `BreakpointValue.kt` in the following test?

src/path_rewriting.rs
```diff
@@ -1045,10 +1047,11 @@ mod tests {
 
     #[cfg(unix)]
     #[test]
-    fn test_rewrite_paths_rewrite_path_for_java_and_rust() {
```
marco-c, line 1048: We should keep a test for java too

src/producer.rs
```diff
@@ -1616,22 +1616,11 @@ mod tests {
...
         let mut file = File::open("./test/jacoco/inner-classes.xml").ok();
         assert!(
             Archive::check_file(file.as_mut(), &Archive::is_jacoco),
             "A Jacoco XML file expected"
         );
-        let mut file = File::open("./test/jacoco/multiple-top-level-classes.xml").ok();
-        assert!(
-            Archive::check_file(file.as_mut(), &Archive::is_jacoco),
-            "A Jacoco XML file expected"
-        );
```
marco-c, line 1634: Why did you remove these?

jasongilanfarr: likely an accident.

marco-c:
@jasongilanfarr thanks for the PR, @AadityaDhingra took it over in #1326, but I'll make sure to reference you in the commit when I merge it.

jasongilanfarr:
Sounds good. Hopefully, i can pull out my changes for all the other
formats. Will also verify against some of my more complex repos with kotlin.

PR #1287, opened 2025-03-22
Accept LLVM profdata files as inputs alongside profraw files

### LINKED ISSUE #1285: Accept *.profdata files as input
## Context

We are using `grcov` to compute coverage in CI in the [uutils/coreutils](https://github.com/uutils/coreutils) and we have a scale problem. Our testsuite produces 1 `.profraw` file (5 MB) for each invocation of the tested binary, which is thousands of times among the whole testsuite.
This creates over 80 GB of trace files, which exceeds the storage capacity of our CI system.

Therefore, I was thinking of splitting the testsuite execution into batches (running the testsuite *per-util*, there is ~100) and to call `grcov` multiple to create several lcov files.
The issue here is that grcov invocations are quite long, and are "duplicating" work, so that's not a viable option either.

## Ideal workflow

Firstly, I'd like to aggregate the batches of `.profraw` files manually into `.profdata` using `llvm-profdata`, mainly because at this point I only need to aggregate the trace files I can delay the creation report after, and I will spare a lot of time.
Then, ideally, I would use grcov and aggregate the `.profdata` using `grcov` and create an lcov report from them.

Would it be possible for grcov to accept profdata files as inputs ? Is there a particular concern that would prevent a trivial implementation of this feature ?

I'm willing to make a contribution if necessary

RenjiSann:
Fixes #1285 

Using `.profdata` files as an input actually proved to be quite straight forward, as when it comes to aggregating them, it's just the same process as for `.profraw` files.
There is no substantial change, apart from dealing correctly with the file extension.

I've decided to keep profraws and profdatas in separate producer archives, though that's discussable.


RenjiSann:
@marco-c if you want to take a look :)

sylvestre:
i was expecting a longer patch!
is that expected you don't have tests ?

RenjiSann:
I haven't figured out how the test framework works just yet, I'll look into it

sylvestre:
@RenjiSann the lint job failed

marco-c (approved):
Thanks!

Issue #1297, opened 2025-03-27
Add option for html output to generate absolute hyperlinks


mightybyte:
I would like to store HTML coverage reports generated by grcov in a non-publicly accessible Google Cloud Storage bucket. When I do this, GCS rewrites the URLs when the page loads. This breaks the coverage report links and makes the report much less useful. It would be nice if there was a way to make the HTML coverage reports generate absolute links with a user-specified prefix that allows the coverage reports to work properly when used in this kind of a link rewriting environment.

PR #1302, opened 2025-03-31
Allow generation of absolute links in HTML output

mightybyte:
Fixes issue #1297


.github/workflows/release.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     tags:
       - v*.*.*
+  workflow_dispatch:
```
marco-c, line 6: I guess this isn't related, right? Could you remove it?

src/html.rs
```diff
@@ -235,11 +235,16 @@ fn get_base(rel_path: &Path) -> String {
 fn get_dirs_result(global: Arc<Mutex<HtmlGlobalStats>>, rel_path: &Path, stats: &HtmlStats) {
     let parent = rel_path.parent().unwrap().to_str().unwrap().to_string();
     let file_name = rel_path.file_name().unwrap().to_str().unwrap().to_string();
+    let mut global = global.lock().unwrap();
     let fs = HtmlFileStats {
         stats: stats.clone(),
+        abs_prefix: global
+            .abs_prefix
+            .as_ref()
+            .map(|p| format!("{}{}/", p, parent)),
```
marco-c, line 244: it'd be ideal, here and elsewhere, to join the abs path and the path with proper formatting and not just as string (otherwise we must ask users to add "/" at the end)

.github/workflows/release.yml
mightybyte: Nope.  It's there so I could test more easily.  I'll remove it once everything else is ready to merge.

src/html.rs
mightybyte: Ahh yes, good point.  I've updated it to use `PathBuf` instead of string concatenation.

src/path_rewriting.rs
```diff
@@ -35,6 +35,12 @@ pub fn canonicalize_path<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
     Ok(path)
 }
 
+//pub fn append_paths(p1: &String, p2: &String) -> io::Result<PathBuf> {
+//    let pb1 = PathBuf::from(p1);
+//    let pb2 = PathBuf::from(p2);
+//    Path::new(p1).join(p2)
+//}
```
marco-c, line 42: WIP

marco-c (changes requested):
LGTM! Please remove the WIP code and then we can merge this

.github/workflows/release.yml
mightybyte: Done

src/path_rewriting.rs
mightybyte: Done

marco-c (approved):

Issue #1079, opened 2023-07-20
Option to use local bulma instead of jsdelivr CDN


legoktm:
Hi, thanks for developing grcov!

On https://doc.wikimedia.org we have a pretty restrictive CSP that only allows local/self-hosted resources for both privacy and security reasons (e.g. [a broken report](https://doc.wikimedia.org/cover/mwbot-rs/mwbot/coverage/)). Would it be possible to bundle any necessary resources with grcov so they don't need to be loaded from a remote site? This would also be nice for offline usage.

See for example how a [PHPUnit report](https://doc.wikimedia.org/cover-extensions/Linter/) bundles bootstrap.

calixteman:
We could just read an env variable which will contain the path of the css file and fallback on the url.
That said the bulma version is pinpointed:
https://github.com/mozilla/grcov/blob/322fc39acacd75aca0ff1c0a1ec2a3e91f04011c/src/html.rs#L75
and we should probably have the url in the Cargo file.

PR #1304, opened 2025-04-03
Add `--html-resources=bundled` to bundle Bulma CSS

pietroalbini:
This PR adds the `--html-resources` CLI flag to control whether to use a bundled version of the CSS and (if added in the future) JS, or to rely on a CDN. When the flag is set to `bundled` (the default), the `BULMA_VERSION` env var will be ignored, while if it's set to `cdn` the old behavior is restored.

Notable things to keep in mind while reviewing this PR:

* https://github.com/mozilla/grcov/pull/1086 already attempted this, but didn't followup on review feedback.

* The bundled Bulma code was taken from the `bulma/css/bulma.min.css` file in the [archive on GitHub releases](https://github.com/jgthms/bulma/releases/download/0.9.1/bulma-0.9.1.zip).
 
* ~~Even though I'd *love* to have this as the default, technically it's a breaking change. So the default behavior is now `--html-resources=cdn` rather than `--html-resources=bundled`. Let me know if I should make the bundled behavior the default.~~ It's now the default.

* Bundling Bulma in grcov means we need to comply with Bulma's license while redistributing grcov. I added the MIT license file in the `src/bulma` directory, ~~and changed Cargo's license field to `MPL-2.0 AND MIT`~~.

Fixes https://github.com/mozilla/grcov/issues/1079


pietroalbini:
Before cutting a new release, keep in mind I have another PR in the works with a significant performance improvement, I'll get to it after lunch :D

Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "grcov"
 version = "0.8.24"
 authors = ["Marco Castelluccio <mcastelluccio@mozilla.com>"]
 description = "Rust tool to collect and aggregate code coverage data for multiple source files"
-license = "MPL-2.0"
+license = "MPL-2.0 AND MIT" # MIT is used for the bundled Bulma CSS
```
marco-c, line 6: We can release the software as MPL-2.0 even if it contains MIT code, so we don't need to modify this.

pietroalbini: Reverted.

pietroalbini:
Addressed review comments! Also, https://github.com/mozilla/grcov/pull/1305 is the other PR I mentioned.

marco-c (approved):

PR #1307, opened 2025-04-04
Speed up binary search

marco-c:
This will help with #1013.

The time to find binaries in the grcov target directory gets 5x faster.

### DIFF
--- a/src/llvm_tools.rs
+++ b/src/llvm_tools.rs
@@ -75,6 +75,28 @@ pub fn run(cmd: impl AsRef<OsStr>, args: &[&OsStr]) -> Result<Vec<u8>, String> {
     Ok(output.stdout)
 }
 
+pub fn find_binaries(binary_path: &Path) -> Vec<PathBuf> {
+    let metadata = fs::metadata(binary_path)
+        .unwrap_or_else(|e| panic!("Failed to open directory '{:?}': {:?}.", binary_path, e));
+
+    if metadata.is_file() {
+        vec![binary_path.to_owned()]
+    } else {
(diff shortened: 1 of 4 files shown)


src/llvm_tools.rs
```diff
@@ -75,6 +68,52 @@ pub fn run(cmd: impl AsRef<OsStr>, args: &[&OsStr]) -> Result<Vec<u8>, String> {
...
+
+            Box::new(move |result| {
+                let entry = result.unwrap();
+
+                if !entry.file_type().unwrap().is_file() {
+                    return Continue;
+                }
+
+                let file = fs::File::open(entry.path()).unwrap();
+                let read = file.take(128).read(&mut bytes).unwrap();
```
calixteman, line 96: It's likely a corner case but what will happen if the file size is lower than 128 ?

marco-c: [take](https://doc.rust-lang.org/std/io/trait.Read.html#method.take) let's us read at most 128 bytes, but if the file is smaller it will stop earlier.
It should be OK even if the file is one byte long: we will overwrite only the first byte in the vector, but then it's unlikely the rest of the bytes will match what is required to detect a binary.
Even in the unlikely case that we read a bit and we are unlucky as we think it's a binary when it isn't, we should still be safe because we ignore the errors when running `export` on the files: https://github.com/mozilla/grcov/blob/09816531636ce8f998a5f86729d34e7d54e380ef/src/llvm_tools.rs#L124-L147.

calixteman (approved):
LGTM. Thank you.

PR #1309, opened 2025-04-04
Slightly reduce allocations when rewriting paths

marco-c:
@pietroalbini could you test this and let me know if it helps in any way?

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -2,7 +2,6 @@ use globset::{Glob, GlobSet, GlobSetBuilder};
 use rayon::prelude::*;
 use rustc_hash::FxHashMap;
 use serde_json::Value;
-use std::collections::hash_map;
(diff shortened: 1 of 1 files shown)


pietroalbini:
The change here was negligible, from 2.8s to 2.770s. After my PR, most of the time is spent inside of walkdir. After 724f64ce5a57af8fe71dc722abf6102ba9e07f0d only ~200 iterations of the loop actually execute code, and reducing allocations for that few iterations is not likely to produce many performance improvements :(

pietroalbini:
Implemented the idea I had over the weekend: https://github.com/mozilla/grcov/pull/1311

marco-c:
I guess it's still worth landing this, even if it helps little, given I already implemented it :P

calixteman (dismissed):
LGTM. Thank you.

PR #1310, opened 2025-04-04
Support providing files and lines to include into report via json file

petrochenkov:
The inclusion set may be determined by some external script, inspecting e.g. git history.

So, here we give some simple json format for such scripts to produce and for grcov to consume.


marco-c:
@petrochenkov could you explain your use case?

PR #1305, opened 2025-04-03
Improve performance of `--source-dir`

pietroalbini:
When trying to run `grcov` in a very large project (containing the Rust repository, all of its submodules, and a large `build/` directory) I noticed that the `--source-dir` flag caused a considerable slowdown:

* `cd path/to/repo && grcov` took 0.06s
* `grcov --source-dir path/to/repo` took 5.4s

It turns out that the code added for https://github.com/mozilla/grcov/pull/196 does a lot of syscalls and allocations when `--source-dir` is passed. In this PR I tried to speed it up a bit, and I got it from 5.4s to 2.8s. Now most of the time is spent in walkdir itself, and improving performance even more would take more time than I have at the moment. Still, I hope the speedup so far is appreciated :D

PR #1311, opened 2025-04-07
Eliminate performance penalty of `--source-dir`

pietroalbini:
Had an idea over the weekend on how to fix the performance issue of `--source-dir`!

After #1305 most of the profile time of `--source-dir` was spent on walkdir iterating through the whole source tree. It wasn't really possible to improve the performance of the walkdir call, as `map_partial_path` requires us to walk through every file in the source directory (to find files named the same as one of the covered files).

The only option left was to not do the walkdir at all, without breaking `map_partial_path`. The idea I had is that calls to `map_partial_path` actually do anything only if one of the covered paths is not a direct child of the source dir. So, if we check that all covered paths exists, we can skip the walkdir altogether!

On my coverage file / source tree, this brought processing time from 2.8s to 0.190s, which is more than acceptable :D

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -246,48 +246,70 @@ pub fn rewrite_paths(
     }
 
     // Traverse source dir and store covered paths, reversed.
+    //
+    // This is a pre-requisite of the `map_partial_path` function, which is needed for Java.
     let mut file_to_paths: FxHashMap<String, Vec<PathBuf>> = FxHashMap::default();
     if let Some(ref source_dir) = source_dir {
-        // When --source-dir points to very large directories (for example, a Rust compiler
-        // checkout with submodules and build artifacts) walking through every file will take a
-        // considerable amount of time. We should skip paths not part of the coverage map.
-        let covered_names = result_map
-            .keys()
-            .map(|path| {
-                path.rsplit_once(['\\', '/'])
-                    .map(|(_dir, file)| file)
-                    .unwrap_or(path.as_str())
-            })
-            .map(OsStr::new)
-            .collect::<HashSet<_>>();
-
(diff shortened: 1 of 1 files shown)


pietroalbini:
I don't think I can spend more time on this than what I did in the past few days :sweat_smile: 

> check if there is any ".java" file in result_map, and only build the map if there is one (or even just make the first call to map_partial_paths fill file_to_paths). This would be an alternative to what you did, and might be even faster (no I/O needed for exists calls);

Yes this could also be a possible alternative! I guess the only footgun would be forgetting to update one place if we change the other place to do the remapping for more than just `.java`.

> fill file_to_paths only for Java paths (this would be beneficial for projects which contain multiple languages).

I don't think this would help, most of the time is spent right now on walkdir listing every file, and we'd still need to do that to find the `.java` files.

marco-c:
> I don't think I can spend more time on this than what I did in the past few days 😅

No problem! Thanks for the contributions so far ;)

> > check if there is any ".java" file in result_map, and only build the map if there is one (or even just make the first call to map_partial_paths fill file_to_paths). This would be an alternative to what you did, and might be even faster (no I/O needed for exists calls);
> 
> Yes this could also be a possible alternative! I guess the only footgun would be forgetting to update one place if we change the other place to do the remapping for more than just `.java`.

We can implement it by defining the list of extensions in a constant that we use everywhere, so we don't run that risk.

> 
> > fill file_to_paths only for Java paths (this would be beneficial for projects which contain multiple languages).
> 
> I don't think this would help, most of the time is spent right now on walkdir listing every file, and we'd still need to do that to find the `.java` files.

Yeah, it would only be a minor improvement to avoid a few allocations.

marco-c (approved):

PR #1314, opened 2025-04-07
Only fill file_to_paths map with Java paths

### LINKED ISSUE #1313: Fill partial path map only for extensions that need it
See the second suggestion from https://github.com/mozilla/grcov/pull/1311#issuecomment-2782894377.

marco-c:
Fixes #1313


calixteman (approved):
LGTM. Thank you.

PR #1321, opened 2025-04-07
Only apply partial path mapping when there are Java files

### LINKED ISSUE #1312: Try to generate partial path mapping only when required
See the first suggestion from https://github.com/mozilla/grcov/pull/1311#issuecomment-2782894377.

marco-c:
Fixes #1312

### DIFF
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -251,21 +251,29 @@ pub fn rewrite_paths(
     //
     // This is a pre-requisite of the `map_partial_path` function, which is needed for Java.
     let mut file_to_paths: FxHashMap<String, Vec<PathBuf>> = FxHashMap::default();
+    let mut map_partial_path_needed = false;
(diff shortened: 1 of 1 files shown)


src/path_rewriting.rs
```diff
@@ -249,21 +249,29 @@ pub fn rewrite_paths(
...
     if let Some(ref source_dir) = source_dir {
         // Calling walkdir over a large directory tree can take a significant time, so we want to
         // execute this only when `map_partial_path` is actually needed.
         //
         // The function's purpose is to figure out whether a covered path is inside a subdirectory.
+
+        // This only happens for Java, so if there are no Java files, we don't need to do it.
+        let has_java = result_map
+            .keys()
+            .any(|path| check_extension(Path::new(&path), "java"));
```
calixteman, line 262: You should use the string you defined in the other patch.

calixteman (dismissed):
LGMT. Thank you.

PR #1322, opened 2025-04-07
Break early as soon as we identify at least one file where mapping partial paths is needed

marco-c:
Depends on #1321


calixteman (approved):
LGTM. Thank you.

PR #1326, opened 2025-04-10
Improve jacoco parsing and improve kotlin support #1280

### LINKED ISSUE #1279: jacoco parsing reports from kotlin

AadityaDhingra:
Rebase and fix review comments in the original PR

Fixes #1279

The jacoco parser made some relatively safe assumptions about the jacoco report, except when jacoco is used to instrument kotlin. Kotlin doesn't have the same restrictions as java when it comes to top level classes, so the report parsing will just immediately error out.

This improves the parser by making it sourcefilename centric instead of class centric (falling back on the pattern of top level class + .java == sourcefilename if the sourcefilename is missing).

In addition, when aggregated across multiple projects, the report commonly doesn't have the full relative paths, so we need to be able to find both java and kotlin files. There is also an assumption that a fully qualified class can only exist in one location and it would panic; however, this isn't entirely uncommon, so don't panic, but instead, inform the user that its not fully supported.


src/parser.rs
```diff
@@ -801,25 +806,12 @@ fn parse_jacoco_report_package<T: BufRead>(
...
-    // relative path.
+    // Change all the keys to use relative paths.
     Ok(results_map
         .into_iter()
         .map(|(class, result)| {
             (
-                format!("{}/{}.java", package, class)
-                    .trim_start_matches('/')
-                    .to_string(),
+                format!("{}/{}", package, class.trim_start_matches('/')).to_string(),
```
marco-c, line 814: I think you can fix the test failure by keeping the old behavior of trimming after format:
```suggestion
                format!("{}/{}", package, class).trim_start_matches('/').to_string(),
```

src/parser.rs
```diff
@@ -801,25 +806,12 @@ fn parse_jacoco_report_package<T: BufRead>(
...
-            return Err(ParserError::InvalidData(format!(
-                "Class {}/{} is not the top class in its file.",
-                package, class
-            )));
-        }
-    }
-
-    // Change all keys from the class name to the file name and turn the result into a Vec.
-    // If package is the empty string, we have to trim the leading '/' in order to obtain a
-    // relative path.
```
marco-c, line 815: And keep this comment

src/path_rewriting.rs
```diff
@@ -337,7 +338,7 @@ pub fn rewrite_paths(
 
             // Try mapping a partial path to a full path.
             let rel_path =
-                if map_partial_path_needed && check_extension(&rel_path, PARTIAL_PATH_EXTENSION) {
+                if map_partial_path_needed && (check_extension(&rel_path, PARTIAL_PATH_EXTENSION) || check_extension(&rel_path, "kt")) {
```
marco-c, line 341: We should make PARTIAL_PATH_EXTENSIONS a vector with both "java" and "kt" and then update the code accordingly

src/parser.rs
```diff
@@ -772,12 +776,13 @@ fn parse_jacoco_report_package<T: BufRead>(
...
-                        let sourcefile = get_xml_attribute(parser, e, "name")?;
-                        let class = sourcefile.trim_end_matches(".java");
+                        // Fully qualified class name: "org/example/Person$Age"
+                        let file = get_xml_attribute(parser, e, "name")?;
+
                         let JacocoReport { lines, branches } =
                             parse_jacoco_report_sourcefile(parser, buf)?;
 
-                        match results_map.entry(class.to_string()) {
+                        match results_map.entry(file.to_string()) {
```
marco-c, line 785: We are indexing by class name here, and by filename in the other case. Maybe this change isn't actually correct and we should keep the old behavior of indexing by class only?

marco-c: Oh I see for Kotlin we have:
```
<class name="BreakpointDirection" sourcefilename="BreakpointValue.kt">
...
</class>
<sourcefile name="BreakpointValue.kt">
```
so it is correct.

Could you check your report from m-c to double check it matches the same format?

AadityaDhingra: yes, we have:

```<package name="mozilla/components/feature/search/suggestions">
<class name="mozilla/components/feature/search/suggestions/SearchSuggestionClientKt" sourcefilename="SearchSuggestionClient.kt"/>
<class name="mozilla/components/feature/search/suggestions/SearchSuggestionClient" sourcefilename="SearchSuggestionClient.kt">
<method name="getStore" desc="()Lmozilla/components/browser/state/store/BrowserStore;" line="30">
<counter type="INSTRUCTION" missed="3" covered="0"/>
<counter type="LINE" missed="1" covered="0"/>
<counter type="COMPLEXITY" missed="1" covered="0"/>
<counter type="METHOD" missed="1" covered="0"/>
</method>```

[whole xml for reference](https://firefoxci.taskcluster-artifacts.net/U0zpZ4Q6SAKzxHeM31nWaQ/0/public/reports/jacoco/jacocoReleaseTestReport/jacocoReleaseTestReport.xml)

marco-c: Perfect!

src/parser.rs
jasongilanfarr: I generally went much more file oriented than package/class oriented. I've since built more (and I'm overdue to open about 10 more PRs) where for example:
- Support go's weird format
- jacoco support for kotlin, scala, and java, including handling of things like missing function / method coverage
- Support for scoverage.xml
- Support for pycov's xml format

A few others. I need to do some work to pull my changes into something manageable as distinct PRs.

marco-c: @jasongilanfarr oh wow, thanks for your work!

marco-c (approved):

PR #1339, opened 2025-04-29
Support go's coverage format

jasongilanfarr:
Support parsing go's coverage format, effectively converting their "Statement Range" coverage into Line Coverage. Branch Coverage isn't available.

This was hand verified against a larger go report by directly comparing go's html to the grcov generated HTML.


src/parser.rs
```diff
@@ -851,6 +852,95 @@ pub fn parse_jacoco_xml_report<T: Read>(
     Ok(results)
 }
 
+pub fn parse_gocov<T: Read>(
+    reader: &mut BufReader<T>,
+) -> Result<Vec<(String, CovResult)>, ParserError> {
+    let re = Regex::new(r"([^:]+):(\d+)\.\d+,(\d+)\.\d+ (\d+) (\d+)").unwrap();
```
marco-c, line 858: nit: we could compile this only once by saving it in a static variable (can be a follow-up)

marco-c (approved):
Thanks!

jasongilanfarr:
I'll do a follow up soon as the current implementation assumes a file
doesn't repeat, but in practice, they do.

Issue #1242, opened 2024-11-05
Don't stop parsing LCOV when hitting an error


marco-c:
Related to #1231

marco-c:
Maybe we should add an option to just log errors vs stop on error, and by default stop on error.

PR #1371, opened 2025-07-08
Allow to ignore lcov parsing error with an argument

a-lubert:
This Pull Request is related to the issue #1242.

Currently when a coverage file contains malformated chunks, the `parse_lcov` function will raised an Error, causing the whole file to be ignored. In some case, this is fine to have the bad part ignored and the rest of the file correctly parse by grcov.

To achieve that, this patch add the argument `--ignore-parsing-error`. Instead of raising an exception when a parsing error occurs, a warning message will be shown and the malformated chuck will be ignored.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -11,7 +11,7 @@ use std::path::Path;
 use std::str;
 use std::sync::Arc;
 
-use log::error;
+use log::{error, warn};
 
 use quick_xml::encoding::Decoder;
 use quick_xml::encoding::EncodingError;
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -205,6 +211,11 @@ pub fn parse_lcov(
                     });
 
                 if key.is_none() {
+                    if ignore_parsing_error {
```
calixteman, line 214: nit: In order to have too much redundant code, I'd use a macro, in order to slightly simplify the code.

src/parser.rs
```diff
@@ -205,6 +211,11 @@ pub fn parse_lcov(
...
+                        parsing_error_occurs = true;
+                        warn!("Invalid key at line {line}, ignore this record");
```
calixteman, line 216: Here and elsewhere in this file, I think it's better to move the iterator to the next EOL (included).
Just in case there is something like `DA:123SF...`

calixteman (approved):
LGTM. Thank you.

PR #1394, opened 2025-08-19
Add multi-threading support for LLVM profile merging

### LINKED ISSUE #1116: grcov --threads parameter is not passed to sub commands
Problem:
grcov executes llvm-profdata and llvm-report if profdata is present, but does not pass on the number of threads to be used. Especially for profdata this slows down llvm-profdata a lot (I have a 50% speed increase from 60 minutes to 90 minutes because of that).

asthetik:
This PR fixes issue #1116 by passing the `--threads` parameter from grcov to the underlying `llvm-profdata merge` command


marco-c:
A couple of things:
1) we should do the same for llvm-cov export;
2) since llvm-cov and llvm-profdata actually do this by default when --num-threads is 0 (and they autodetect the right number of threads to use), if the grcov parameter is not passed explicitly by the user, let's rely on their autodetection instead of ours.

asthetik:
Hi @marco-c , I have updated the changes you mentioned, thanks.

src/lib.rs
```diff
@@ -179,6 +179,7 @@ pub fn consumer(
     guess_directory: bool,
     binary_path: Option<&Path>,
     ignore_parsing_error: bool,
+    llvm_threads: Option<usize>,
```
marco-c: I'd call this num_threads

src/main.rs
```diff
@@ -465,6 +466,7 @@ fn main() {
                     guess_directory,
                     binary_path.as_deref(),
                     ignore_parsing_error,
+                    llvm_threads,
```
marco-c: Just pass opt.threads and avoid defining an intermediate variable

asthetik:
@marco-c All your suggestions are now addressed, and all tests pass. The CI lint failure is unrelated to this pr.

marco-c (approved):
Thanks!

PR #1500, opened 2026-05-21
fix: handle optional DA checksum field in LCOV parser

robertomlsoares:
The LCOV format allows an optional checksum field on DA lines: `DA:line_number,execution_count[,checksum]`

When a checksum starts with `e`, the byte-level parser mistakes it for an `end_of_record` marker, since it only checks the first byte. This causes a panic at `parser.rs:194` (unwrap on `None`).

The fix validates the full `end_of_record` string instead of matching on just the `e` byte. Lines starting with `e` that don't match `end_of_record` are skipped.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -185,6 +185,13 @@ pub fn parse_lcov(
         line += 1;
         match *c {
             b'e' => {
+                // Verify this is actually "end_of_record" and not a
+                // stray 'e' (e.g. from an optional DA checksum field).
+                let rest: Vec<u8> = iter.take_while(|&&c| c != b'\n').copied().collect();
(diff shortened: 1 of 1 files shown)


robertomlsoares:
Looks like the linter is failing on pre-existing code.

robertomlsoares:
I have tried this fix in our systems, and while it did work for checksums starting with `e`, it made me realise this issue is much more widespread. It affects all checksums that have starting characters equal to some LCOV "keywords" such as `FN`, `DA`, `SF`, etc. The appropriate fix is much more complex than what this PR is currently doing, so I will close it and leave the problem documented here https://github.com/mozilla/grcov/issues/1501.

PR #1513, opened 2026-07-03
feat: Prevent raising error for gcov format v2

### LINKED ISSUE #1512: [Feature request] gcov format version 2 support
GCC 14 bumped its `gcov` intermediate JSON schema to `"format_version": "2"`. And using `grcov` together with it generates this error:

```
[ERROR] Format version 2 is not expected, please file a bug on https://github.com/mozilla/grcov
```

This is triggered in [`src/parser.rs`](https://github.com/mozilla/grcov/blob/master/src/parser.rs):

```rust
if gcov.format_version != "1" {
    error!(
        "Format version {} is not expected, please file a bug on https://github.com/mozilla/grcov",
        gcov.format_version
    );
}
```

Currently `grcov` still generates coverage report which seems to be correct despite of this error message. But I am not very sure whether the best fix is to remove this check directly.

Reproduction (has a dockerfile so we have gcc14 environment):

[grcov-gcc14-repro.tar.gz](https://github.com/user-attachments/files/29344934/grcov-gcc14-repro.tar.gz)

fangsong-cellink:
Closes #1512

According to[ `gcov` document](https://gcc.gnu.org/onlinedocs/gcc-14.1.0/gcc/Invoking-Gcov.html):

> format_version: semantic version of the format
> Changes in version 2:
>         calls: information about function calls is added 

the change is purely additive.

So it is safe to just relax the version check a bit.

This PR also fixes linting errors, even if they are not caused by my change


tchernobog:
@marco-c this would silence quite a few warnings for us, any drawbacks in merging it?

marco-c (dismissed):

marco-c:
Can you add a v2 file as a test case for this?

fangsong-cellink:
@marco-c Test file added.

Generated with:
```cpp
#include <cstdio>

int uncovered(int a) {
  if (a > 0) {
    return a * 2;
  }
  return -a;
}

int covered(int a) {
  if (a % 2 == 0) {
    return a / 2;
  } else {
    return a * 3 + 1;
  }
}

int main() {
  int total = 0;
  for (int i = 0; i < 4; i++) {
    total += covered(i);
  }
  printf("%d\n", total);
  return 0;
}
```

```sh
g++-14 --coverage -O0 -o gcov_format_v2 gcov_format_v2.cpp
./gcov_format_v2                              
gcov-14 -b -c gcov_format_v2.gcno -i 
```

marco-c (approved):
Thanks!

PR #1515, opened 2026-08-06
fix(parser): stop panicking on malformed (and on some valid) lcov input

### LINKED ISSUE #1501: LCOV parser breaks on DA lines with optional checksum field
The LCOV format allows an optional checksum field on DA lines: `DA:line_number,execution_count[,checksum]`

grcov's byte-level LCOV parser does not consume the optional checksum field after reading the execution count. Depending on what character the checksum starts with, this causes different issues.

1. checksum starts with `e` (interpreted as `end_of_record`):
```
SF:test.py
DA:1,0,e3ZBlmsTAEA3QwKvJLg9CA
LH:0
LF:1
end_of_record
A panic occurred at src/parser.rs:194: called `Option::unwrap()` on a `None` value
```

2. checksum starts with a valid key prefix like `FN`, `DA`, `FNDA`:
```
SF:test.py
DA:1,0,FNsomechecksum
LH:0
LF:1
end_of_record
Error parsing file: Invalid record: 'FN at line 3'
```

3. checksum starts with `SF` or `BRDA`:
```
SF:test.py
DA:1,0,SFsomechecksum
LH:0
LF:1
end_of_record
No error, but the checksum is interpreted as a new source file record, silently corrupting the output.
```

I have tried tackling the issue here https://github.com/mozilla/grcov/pull/1500 before realising it is more widespread, and not isolated to just checksums starting with `e`.

### LINKED ISSUE #1511: `grcov` can panic on several malformed coverage inputs instead of reporting parse/output errors
I found a few robustness issues where malformed coverage artifacts or path data can make `grcov` panic during parsing/report generation.

The main cases below are reachable through normal coverage-processing inputs: LCOV text, GCNO binary data, and filesystem paths discovered by the producer/HTML pipeline. A single bad artifact can abort the current report run instead of being handled as a parse/output error.

I checked the nearby comments around the panic sites below. The comments describe record formats or output behavior, but I did not find comments or docs that cover these panic preconditions.

Version checked: `grcov 0.10.7`

## Main examples

| Area | Panic site | Trigger |
| --- | --- | --- |
| LCOV parser | `src/parser.rs:259`, `src/parser.rs:275`, `src/parser.rs:281`, `src/parser.rs:345`, `src/parser.rs:385`, `src/parser.rs:395`, `src/parser.rs:405` | Oversized numeric fields or duplicate counters overflow during manual decimal parsing/aggregation. |
| LCOV parser | `src/parser.rs:194`, `src/parser.rs:646` | `end_of_record` or final lines before any `SF:` record make `cur_file.unwrap()` panic. |
| GCNO parser | `src/reader.rs:245` | A non-empty all-zero string field makes `position(...).unwrap()` panic. |
| Path handling | `src/producer.rs:33`, `src/producer.rs:68`, `src/html.rs:206`, `src/html.rs:428` | Non-UTF-8 filenames/extensions or malformed relative paths are unwrapped during producer/HTML output. |

## Example 1: LCOV numeric overflow

Relevant code:

```rust
let line_no = iter
    .take_while(|&&c| c.is_ascii_digit())
    .fold(0, |r, &x| r * 10 + u32::from(x - b'0'));

let execution_count = iter
    .take_while(|&&c| c.is_ascii_digit())
    .fold(u64::from(*c - b'0'), |r, &x| {
        r * 10 + u64::from(x - b'0')
    });

*cur_lines.entry(line_no).or_insert(0) += execution_count;
```

The parser already has a `ParserError` path for malformed records, but very large decimal fields can overflow before that path is used.

Minimal reproducer shape:

```rust
use grcov::parse_lcov;

#[test]
#[should_panic]
fn lcov_da_counter_overflow_panics() {
    let input = b"SF:/tmp/overflow.rs\nDA:7,18446744073709551615\nDA:7,1\nend_of_record\n".to_vec();

    let _ = parse_lcov(input, false, false);
}
```

Other LCOV fields use the same unchecked decimal-fold pattern, including `FNDA` and `BRDA`.

Expected behavior:

Malformed or oversized LCOV numbers should return `ParserError` or be ignored when `ignore_parsing_error` is enabled.

Actual behavior:

Debug/test builds can panic on integer overflow. Release builds may silently wrap and continue with incorrect coverage data.

## Example 2: LCOV record before `SF:` panics

Relevant code:

```rust
// we've a end_of_record
results.push((
    cur_file.unwrap(),
    CovResult {
        lines: cur_lines,
        branches: cur_branches,
        functions: cur_functions,
    },
));
```

Minimal reproducer shape:

```rust
use grcov::parse_lcov;

#[test]
#[should_panic]
fn lcov_end_record_without_source_file_panics() {
    let input = b"end_of_record\n".to_vec();

    let _ = parse_lcov(input, false, false);
}
```

Expected behavior:

An LCOV record without a source file should be rejected as malformed input.

Actual behavior:

The parser unwraps the missing source file and panics.

## Example 3: GCNO all-zero string field

Relevant code:

```rust
let bytes = &self.buffer[start..self.pos];
let i = len - bytes.iter().rev().position(|&x| x != 0).unwrap();
Ok(unsafe { std::str::from_utf8_unchecked(&bytes[..i]).to_string() })
```

If the GCNO string field has a non-zero word length but all bytes are zero, `position(|&x| x != 0)` returns `None`.

Minimal reproducer shape:

```rust
use grcov::Gcno;

#[test]
fn gcno_all_zero_string_field_panics() {
    let mut gcno = Vec::new();

    gcno.extend_from_slice(b"oncg");               // little-endian gcno marker
    gcno.extend_from_slice(b"*09A");               // version that reaches read_string
    gcno.extend_from_slice(&0u32.to_le_bytes());   // checksum
    gcno.extend_from_slice(&1u32.to_le_bytes());   // string length: one u32 word
    gcno.extend_from_slice(&[0, 0, 0, 0]);          // all-zero string payload

    let result = std::panic::catch_unwind(|| {
        let _ = Gcno::compute("bad.gcno", gcno, vec![], false);
    });

    assert!(result.is_err());
}
```

Expected behavior:

The malformed GCNO string should produce a structured parse error.

Actual behavior:

The parser panics on `unwrap()`.

## Example 4: non-UTF-8 filenames during producer/HTML handling

Relevant code:

```rust
fn clean_path(path: &Path) -> String {
    path.to_str().unwrap().to_string()
}

if let Some(ext) = path.extension() {
    match ext.to_str().unwrap() {
        "gcno" => { /* ... */ }
        "gcda" => { /* ... */ }
        "info" => { /* ... */ }
        _ => {}
    }
}
```

HTML output has the same pattern when building output names:

```rust
let mut ext = ext.to_str().unwrap().to_owned();
let filename = rel_path.file_name().unwrap().to_str().unwrap();
```

Minimal reproducer shape on Unix:

```rust
use crossbeam_channel::unbounded;
use grcov::producer;
use std::ffi::OsString;
use std::fs;
use std::os::unix::ffi::OsStringExt;

#[test]
#[should_panic]
fn non_utf8_info_filename_panics_in_producer() {
    let base = tempfile::tempdir().unwrap();

    let mut bad_name = vec![b'f', b'o', 0x80, b'o'];
    bad_name.extend_from_slice(b".info");
    fs::write(base.path().join(OsString::from_vec(bad_name)), b"TN:\n").unwrap();

    let (sender, _receiver) = unbounded();
    let inputs = vec![base.path().to_string_lossy().into_owned()];

    let _ = producer(base.path(), &inputs, &sender, false, false);
}
```

Expected behavior:

The producer/report pipeline should skip the malformed path, preserve it as an `OsStr`, or return/log a normal error.

Actual behavior:

The path is converted with `to_str().unwrap()` and the report run panics.

## Suggested fix direction

- Use checked decimal parsing for LCOV numeric fields, and route failures through `ParserError`.
- Treat missing `SF:` as a malformed LCOV record instead of unwrapping `cur_file`.
- In GCNO string parsing, handle all-zero payloads without unwrapping `None`.
- Avoid `Path::to_str().unwrap()` for discovered files; either use lossy display only for messages or propagate a recoverable path error.

<details>
<summary>Additional lower-priority observations</summary>

### Coverage-stat aggregation helpers

There are several public or semi-public aggregation helpers that assume valid internal counts:

```rust
pub fn new(total: usize, covered: usize, precision: usize) -> Self {
    let missed = total - covered;
    /* ... */
}

pub fn add(&mut self, other: &Self) {
    self.total += other.total;
    self.covered += other.covered;
    self.missed += other.missed;
}

if let Some(line) = lines.get_mut((*line_num - 1) as usize) {
    /* ... */
}
```

These can panic for inputs such as `covered > total`, very large accumulated counters, extreme precision, or line number `0`. I would mention them as hardening opportunities, but I would not lead with them unless the maintainers consider those constructors part of the supported public API.

### GCOV version/tool handling

`get_gcov_version()` and `parse_version()` assume `gcov --version` succeeds, emits UTF-8, and contains a parseable semantic version:

```rust
let output = Command::new(get_gcov())
    .arg("--version")
    .output()
    .expect("Failed to execute `gcov`. `gcov` is required (it is part of GCC).");
assert!(output.status.success(), "`gcov` failed to execute.");
let output = String::from_utf8(output.stdout).unwrap();
assert!(version.is_some(), "no version found for `gcov`.");
```

This is probably best framed as CLI/tooling robustness, not as a core parser bug.

### Template, channel, mutex, and writer failures

Generated tests also found panics around missing templates, closed work channels, poisoned mutexes, and output writer failures. Those are real panic paths, but they are weaker as issue lead examples because they depend more on internal workflow state or injected failing components.

</details>

Eljees:
Fixes #1511. Also fixes #1501.

`parse_lcov` has a dedicated error path — the `manage_parsing_error!` macro and `ParserError::InvalidRecord` — but two `unwrap()` calls sitting next to it panic instead of using it. With lenient parsing (`ignore_parsing_error = true`) the panic happens even though the caller explicitly asked the parser not to fail.

**1. Any line starting with `e` was treated as `end_of_record`.**
The outer dispatch (`src/parser.rs:186`) matches on the *first byte* of the line, so `excluded_by_a_tool` — or any other record grcov does not know — took the `end_of_record` branch and hit `cur_file.unwrap()` on line 194. This is reachable from **valid** lcov files.

The same byte-level dispatch is why `DA:<line>,<count>,<checksum>` (the optional third field of a `DA` record, standard lcov) crashes: the `DA` branch stops after `<count>`, the checksum `e4f2` is then read as the start of a new line, and `e` → panic. That is issue #1501.

**2. `end_of_record` before any `SF:` panicked** instead of reporting `InvalidRecord`.

**3. An unknown key longer than four uppercase letters panicked in lenient mode.**
`key` is already computed with `checked_mul`/`checked_add` and `manage_parsing_error!` is already called when it overflows — but in lenient mode the macro only logs, and the following `match key.unwrap()` (line 231) panics anyway.

The patch routes all three through the existing error path and makes the `end_of_record` branch check the whole token instead of one byte. No new error semantics are introduced: unknown records are skipped exactly like the ones already handled by the catch-all arm.

Six unit tests are added, in the style of the existing inline-buffer tests (`test_lcov_parser_empty_DA_record`). Five of them panic on `master`.

<details><summary>before / after</summary>

```
$ cargo test --lib                # master
test result: FAILED. 133 passed; 4 failed        # 4 = llvm_tools::*, need llvm-profdata

$ cargo test --lib lcov_parser    # master + new tests only
test result: FAILED. 8 passed; 5 failed
  ... panicked at src/parser.rs:194:30: called `Option::unwrap()` on a `None` value
  ... panicked at src/parser.rs:231:27: called `Option::unwrap()` on a `None` value

$ cargo test --lib                # master + new tests + fix
test result: FAILED. 139 passed; 4 failed        # same 4 llvm_tools failures as on master
```
</details>

**Not in this PR, tracked for a follow-up.** #1511 also mentions the hand-rolled decimal folds. They are genuinely reachable, but only from input no producer emits, and fixing them touches seven call sites, so I kept the diff focused. Reproducers, all panicking on `master` today:

```
DA:7,18446744073709551615   ->  src/parser.rs:281  attempt to add with overflow
DA:99999999999,1            ->  src/parser.rs:259  attempt to multiply with overflow
DA:5,!                      ->  src/parser.rs:274  attempt to subtract with overflow
```

Happy to send that as a second PR (or fold it in here) if you prefer.

*AI-assisted: I used Claude to help map the panic sites and to draft the patch and the tests. I reviewed every line, ran each new test against unpatched `master` first to confirm it panics without the fix, and ran `cargo fmt --check` and `cargo clippy -- -D warnings` per the repo's pre-commit config; the analysis and the runs are mine.*


Eljees:
A note on the red `Lint` job, since it is not caused by this PR.

All five clippy errors are outside the diff:

```
src/producer.rs:112:21        this `if` can be collapsed into the outer `match`
src/producer.rs:148:26        can be more succinctly written as a byte str
src/producer.rs:148:57        can be more succinctly written as a byte str
src/path_rewriting.rs:426:21  this `if` can be collapsed into the outer `match`
src/cobertura.rs:67:45        iterating on a map's values
```

This branch only touches `src/parser.rs`. I also ran the pre-commit clippy hook in a container against both trees — pristine `master` and `master` plus this patch — and on the toolchain I had available (1.88) each exits 0, which suggests the lints above were added to clippy some time after the last commit on `master` rather than by anything here. All ten test jobs on this PR are green.

If it would help, I am happy to send those five as a separate cleanup PR to get `Lint` back to green; I kept them out of this one so the diff stays on the parser.

src/parser.rs
```diff
@@ -189,21 +189,40 @@ pub fn parse_lcov(
                     continue;
                 }
 
+                // Only `end_of_record` closes a record: any other line starting
+                // with 'e' is an unknown record and has to be skipped like the
+                // ones handled by the catch-all arm below.
+                let mut rest: Vec<u8> = iter.take_while(|&&c| c != b'\n').copied().collect();
```
calixteman, line 195: Can we avoid to copy the bytes here ? it's useless.

Eljees: `rest` isn't dead code here — it's compared against `b"nd_of_record"` (with an optional trailing `\r` stripped) to detect the record terminator, so the check itself is needed. But you're right that collecting it into a `Vec` just for that comparison is unnecessary; I'll rework it to compare against the expected bytes without allocating.

Eljees:
Updated the branch onto current `master`, and the red `Lint` I flagged above is gone: the five clippy errors I listed were all outside this diff and `master` has since dealt with them. The run on the new head is green.

Nothing else changed - the diff is the same one, still the parser fix for malformed lcov records plus its tests.

Issue #1514, opened 2026-07-09
Parser rejects `FN:` records with negative line numbers, then deadlocks when panics happen in worker threads


gergelyfabian:
## Summary

`grcov`'s lcov parser rejects `FN:<line>,<name>` records where `<line>` is negative (e.g. `FN:-1,...`) and **discards the entire input file's coverage data** as a result. When many worker threads panic in parallel during multi-file processing, the main thread also `futex_wait`s indefinitely on results the panicked workers never produce, and `grcov` hangs at ~0 % CPU with no exit and no useful output.

JaCoCo emits `FN:-1,...` for synthetic/generated methods that lack a source line — Scala auto-generated `.curried` / `.tupled` on case classes, macro-expanded methods, etc. — so any medium-sized Scala/JVM project routes through this bug the first time `grcov` merges its coverage output.

## Environment

- `grcov 0.10.7`, `grcov-x86_64-unknown-linux-musl` from the GitHub release.
- Input: lcov `.info` files emitted by Bazel `bazel coverage --combined_report=lcov` for a Scala/JVM codebase (JaCoCo → lcov converter).

## Minimal reproducer — parse fault / data loss

Save as `in/test.info`:

```
TN:
SF:foo.scala
FN:-1,curried
FNDA:0,curried
FNF:1
FNH:0
DA:1,0
LF:1
LH:0
end_of_record
```

Run:

```
grcov in -t lcov -o out.info
```

### Actual output

`grcov` stderr:

```
[ERROR] Error parsing file /path/to/in: Invalid record: 'FN at line 3'
```

`out.info` contains only `TN:` — the file's entire coverage data is silently discarded.

### Expected

Either accept negative line numbers (parse as `i32` / treat as "synthetic, no source location"), or skip only the malformed record and preserve the rest of the file's coverage.

## Empirical narrowing

Same input with only the FN's line number changed:

| `FN:` value | Result |
|---|---|
| `FN:-1,curried`     | FAIL — whole file dropped |
| `FN:-999,curried`   | FAIL — same |
| `FN:-1,name (I)Z`   | FAIL — same (space in name is irrelevant) |
| `FN:0,curried`      | OK  |
| `FN:1,curried`      | OK  |
| `FN:1,name (I)Z`    | OK — space in JVM signature is fine |
| `FN:,curried`       | FAIL (expected — empty line number) |

[...]

PR #1516, opened 2026-08-07
fix(parser): accept FN records with a negative line number

Eljees:
Fixes the parsing half of #1514.

`FN:-1,<name>` is emitted for synthetic functions that have no source line — the reporter hit it with JaCoCo output for Scala's `.curried`/`.tupled`. The `FN` branch peeks at the first byte and rejects anything that is not a digit (`src/parser.rs:289`), so with the default (strict) mode `manage_parsing_error!` returns `Err` and **the whole coverage file is dropped**, not just the one record.

`Function.start` is a `u32`, so a negative line cannot be represented. The patch accepts an optional leading `-`, treats such a function as synthetic and records it at line 0, which keeps the function name and its `FNDA` counter working. A missing line number (`FN:,name`) stays an error — there is a test for that.

<details><summary>before / after</summary>

```
$ cargo test --lib FN_record    # master + new tests only
test test_lcov_parser_FN_record_without_line_number ... ok        # control
test test_lcov_parser_FN_record_with_negative_line_number ... FAILED
  ... called `Result::unwrap()` on an `Err` value: InvalidRecord("FN at line 3")

$ cargo test --lib              # master + new tests + fix
test result: FAILED. 135 passed; 4 failed        # same 4 llvm_tools failures as on master
```
</details>

**The deadlock half of #1514 is not addressed here.** The reporter also describes the main thread hanging in `futex_wait` when many worker threads panic. That is a change to the threading model in `src/main.rs`, and the trigger is the panics themselves — several of which are removed by #1515. Worth tracking separately.

**Relationship to #1515.** Both branches touch `src/parser.rs`, but they are independent and each is based on `master`, so they can be reviewed in either order. If #1515 lands first this one needs a trivial rebase — say the word and I will push it.

*AI-assisted: I used Claude to help locate the strict-mode rejection and to draft the patch and the tests. I reviewed every line, ran the new test against unpatched `master` first to confirm it fails without the fix, and ran `cargo fmt --check` and `cargo clippy -- -D warnings` per the repo's pre-commit config; the analysis and the runs are mine.*


Eljees:
Same note as on #1515 about the red `Lint` job: it is pre-existing and not caused by this PR.

The five clippy errors are in `src/producer.rs`, `src/path_rewriting.rs` and `src/cobertura.rs`; this branch only touches `src/parser.rs`. Running the pre-commit clippy hook against pristine `master` and against `master` plus this patch gives exit 0 in both cases on the toolchain I had available, so the CI failure comes from a newer clippy rather than from the change. All ten test jobs here are green.

I listed the exact lints and offered a separate cleanup PR over on #1515, to keep the discussion in one place.

src/parser.rs
```diff
@@ -286,6 +286,14 @@ pub fn parse_lcov(
...
 
                         // FN:int,string
+                        // A negative line number marks a synthetic function
+                        // that has no location in the source file: keep the
+                        // record and report it at line 0 instead of discarding
+                        // the whole coverage file.
+                        let synthetic = iter.peek() == Some(&&b'-');
+                        if synthetic {
+                            iter.next();
+                        }
```
marco-c, line 312: Have you had a look at what other tools do? Do they also pretend these are at line 0?

src/parser.rs
```diff
@@ -298,6 +306,7 @@ pub fn parse_lcov(
                         let start = iter
                             .take_while(|&&c| c.is_ascii_digit())
                             .fold(0, |r, &x| r * 10 + u32::from(x - b'0'));
+                        let start = if synthetic { 0 } else { start };
```
calixteman, line 325: If the other tools handle negative numbers as 0, then we should just skip number parsing: it's useless.

Eljees: Still checking this properly, but so far: lcov's own `BRDA` records follow the same pattern — exit branches with a negative line number get normalized to 0 to stay format-valid — and there's a related discussion on the gcov/C++ side about artificial functions in coverage reports (https://discourse.llvm.org/t/incorrect-line-coverage-caused-by-artificial-functions-in-gcov-coverage-reports/84732). Haven't run this through `genhtml` myself yet to be sure it doesn't choke on it.

@calixteman agreed on the parsing — right now the fold still runs for synthetic records and gets thrown away right after. I'll keep the `take_while` (still need to consume the digits to stay positioned for the comma that follows) but skip the fold itself when `synthetic` is set.

Eljees:
Not exhaustively, and I don't want to overstate what I found. I checked `genhtml`'s own
source for special-casing of negative/zero-line `FN` records and found none — lcov's HTML
renderer doesn't appear to encode a documented convention either way for this, so that's not
a precedent I can point to.

The closer analogy is on the producer side: negative-line `FN` records for synthetic
functions are the same class of problem as artificial/compiler-generated functions
contaminating gcov-based coverage more generally, e.g. `__cxx_global_var_init()` and similar
in the C++/gcov world — discussed here:
https://discourse.llvm.org/t/incorrect-line-coverage-caused-by-artificial-functions-in-gcov-coverage-reports/84732

I haven't found a specific tool that documents "map it to line 0" as the answer. If you'd
rather this PR represent "no known line" differently (e.g. skip reporting a line for these,
or keep a sentinel other than 0) I'm open to that — 0 was just the smallest change that keeps
the record from being dropped. Let me know which you'd prefer and I'll adjust.

Eljees:
Correcting my note above: I had looked in `bin/genhtml`, which is the wrong file. lcov's `.info` reader lives in `lib/lcovutil.pm`, and there is a convention there after all.

**The `FN:` line number is matched with `\d+` in every version I checked, so a leading `-` is never accepted** and `FN:-1,<name>` is not a valid `FN` record for lcov:

* v1.16, `bin/lcov` [L1730](https://github.com/linux-test-project/lcov/blob/v1.16/bin/lcov#L1730): `/^FN:(\d+),([^,]+)/`
* v2.0, `lib/lcovutil.pm` [L4694](https://github.com/linux-test-project/lcov/blob/v2.0/lib/lcovutil.pm#L4694): `/^FN:(\d+),((\d+),)?(.+)$/`
* v2.5, `lib/lcovutil.pm` [L9154](https://github.com/linux-test-project/lcov/blob/v2.5/lib/lcovutil.pm#L9154): same

With no branch matching, the record falls through to the catch-all, which raises `ERROR_FORMAT` — `unexpected .info file record` ([v2.5 L9409-L9410](https://github.com/linux-test-project/lcov/blob/v2.5/lib/lcovutil.pm#L9409-L9410)).

**Line 0 is not an ordinary location either, in current lcov.** v2.5 checks for it by name: `$lineNo <= 0` raises `ERROR_FORMAT` with `unexpected function line '0'` ([L9160-L9171](https://github.com/linux-test-project/lcov/blob/v2.5/lib/lcovutil.pm#L9160-L9171)). That check is absent in v2.0, so older lcov accepts `FN:0` silently; it was added somewhere between the two.

**How the error behaves.** Both paths go through `ignorable_error` ([L2332-L2369](https://github.com/linux-test-project/lcov/blob/v2.5/lib/lcovutil.pm#L2332-L2369)): fatal by default, downgraded to a warning under `--ignore-errors format`. In the warning case parsing continues, and on the `<= 0` path `define_function` still runs right after ([L9174-L9176](https://github.com/linux-test-project/lcov/blob/v2.5/lib/lcovutil.pm#L9174-L9176)), so the function survives with the line as given.

So, directly:

**@marco-c** — no, lcov does not quietly place these at line 0. It rejects a negative line as a malformed record, and current lcov rejects a zero line too. The difference from grcov is granularity rather than strictness: lcov's failure is per record and suppressible, and suppressing it keeps the rest of the file, while grcov's default mode fails the whole file.

**@calixteman** — that changes my own view of this patch. Since 0 is a value lcov itself flags, having grcov substitute it silently is inventing a location. Two options, both smaller than what I pushed:

1. Treat `FN:-<n>` as a malformed record and skip just that record, keeping the file. This is closest to lcov under `--ignore-errors format`, adds no new semantics, and it fixes the symptom in #1514 — the whole file being dropped. The function is lost from the report.
2. Keep the record at line 0 as the current patch does, but warn, so the report says the location is unknown instead of pointing at line 0 as though it were real.

I lean to (1) unless you would rather the function survived in the report. Say which and I will rework the branch.

For transparency: the above is read from the released sources at the tags linked, and I have not run lcov end to end on such a file.

Eljees:
I said I had not run it; I have now, and the run adds something that changes which of the two options I would pick.

Environment: lcov `2.0-1` as packaged in Ubuntu 24.04. Three otherwise identical `.info` files differing only in the `FN` line, each with a matching `FNDA:0,curried`:

```
FN:2,curried    -> accepted; lcov --summary and genhtml both report functions..: 0.0% (0 of 1)
FN:0,curried    -> accepted silently, no warning at all; same 0 of 1 function
FN:-1,curried   -> rejected
```

The rejection is the catch-all path, verbatim:

```
$ lcov --summary neg.info
lcov: ERROR: "neg.info":3: unexpected .info file record 'FN:-1,curried'
	(use "lcov --ignore-errors format ..." to bypass this error)

$ genhtml -o out neg.info
genhtml: ERROR: "neg.info":3: unexpected .info file record 'FN:-1,curried'
	(use "genhtml --ignore-errors format ..." to bypass this error)
```

Two things worth noting.

**`FN:0` really is silent on 2.0.** That confirms the `$lineNo <= 0` check I linked is a later addition and not something 2.0 enforces, so "line 0 is rejected" holds for current lcov only.

**Skipping the record is not free** — this is the part I had not anticipated:

```
$ genhtml --ignore-errors format -o out neg.info
genhtml: WARNING: ('format') "neg.info":3: unexpected .info file record 'FN:-1,curried'
genhtml: ERROR: unknown function 'curried'
	(use "genhtml --ignore-errors mismatch ..." to bypass this error)
```

Dropping the `FN` record leaves its `FNDA` counter pointing at a function that was never defined, and lcov raises that as a *second*, separately-suppressible error. So option (1) from my previous comment is not a one-line skip: whatever drops the `FN` record has to drop the paired `FNDA` too, or the file still fails for a different reason.

Given that, I would now go with option (2): keep the function, record it at line 0, and emit a warning so the report does not present line 0 as a real location. It keeps `FN`/`FNDA` consistent, it is what lcov 2.0 itself does for `FN:0`, and it stays a small change to `src/parser.rs`. @calixteman, on your point about skipping the number parse — I agree the parsed digits are unused on that path, so I would consume them without folding rather than compute a value and discard it.

Say the word and I will push that revision, warning included.

marco-c:
We have a mode to skip errors (https://github.com/mozilla/grcov/blob/c9185272655bcd2c2d346ebda586b78f84899b6d/src/main.rs#L204), we could do 1 when this mode is enabled?

Eljees:
That's a clean way to settle this without pinning down what other tools do — makes sense to me.

Gated it on `--ignore-parsing-error`: a negative `FN` line number is now a hard parse error (exit 1) when that flag is off, same as any other malformed record. With the flag on, it's normalized to line 0 and parsing continues (exit 0), same as before. Added tests for both paths, ran the full suite (138 tests) and clippy clean locally. Pushed as 99800d0, rebased on current master.

marco-c (approved):

PR #1517, opened 2026-08-12
Support the GCC 12.1 and newer gcov format in the built-in reader

michieldwitte:
The built-in gcno/gcda reader only understood the format up to GCC 11, so anything newer failed to parse and coverage had to go through gcov instead. On GCC 15.2 output, master errors with Not enough data in buffer: cannot skip 10804527104 bytes and reports nothing at all.

GCC 12.1 changed three things: the gcno and gcda header carry a checksum after the stamp, lengths count bytes instead of 4-byte words and strings lost their padding (so records are no longer 4-byte aligned), and a negated record length marks counters that are all zero and left out. The reader now handles all three, keyed off the version it parsed, so older files are read exactly as before. Since records can now be unaligned, integers are read from their bytes rather than through a  transmuted pointer.

Tests: gcc-11, gcc-12 and gcc-15 reader fixtures in the existing style plus one for the omitted zero counters, and version assertions on both sides of the 12.1 boundary. The golden dumps for gcc-9 through gcc-15 are byte-identical, so the same program parses to the same model across the format change. cargo fmt --check and clippy are clean, and 139 tests pass — the 4 llvm_tools failures want llvm-profdata installed locally and fail the same way on master.

On a real C++ project (GCC 15.2, coverage from 174 separate test runs) the built-in reader reports the same 64.8% whether the runs are merged with gcov-tool merge first or passed in unmerged.

Depends on #1524: adding gcc-12 and gcc-15 gcno to test/ makes the two repo-wide test_coveralls_* tests run gcov 15 over gcno of other majors, and the leftover json of those failed runs gets read as the next gcno's output. That is the gcc-15 CI failure here, reproducible locally only when both tests run concurrently. With #1524 in, the row passes.


marco-c:
@michieldwitte could you split this into multiple self-contained PRs?

michieldwitte:
> @michieldwitte could you split this into multiple self-contained PRs?

Done, this PR is now just the built-in reader format support.  
(It does need #1524 for the tests to succeed)
The other PR's:  
Read the gcda of a merged gcno in parallel - #1523 
Don't crash on the gcov output of an unsupported gcno - #1524 
One more is waiting on this one: skipping a gcno whose GCC major the local gcov cannot read, which takes the repo-wide test_coveralls_* tests from 157s to 1.9s with gcov 15.  
Its test reuses the gcc-11/12/15 fixtures added here, so I'll open it once this lands rather than duplicate them.

calixteman:
@michieldwitte can you rebase ?

PR #1518, opened 2026-08-27
chore: satisfy clippy on the current toolchain

Eljees:
`Lint` fails on `master` itself with the clippy CI now installs, so it is red on every pull request opened since that bump - 13 of the 25 open ones when I checked, including 11 dependabot bumps, #1394, and my own #1515 and #1516. None of the findings are in code those branches touch, so nobody can turn their own PR green.

The five findings CI reports, and what this does with each:

| lint | site | change |
| --- | --- | --- |
| `iter_kv_map` | `cobertura.rs:67` | `into_iter().filter_map(\|(_, l)\| ...)` discards the key, so it is `into_values()` |
| `collapsible_match` | `path_rewriting.rs:426` | the two `Some(..)` arms fold into match guards |
| `byte_char_slices` x2 | `producer.rs:148` | `[b'T', b'N', b':']` is `*b"TN:"` |
| `collapsible_match` | `producer.rs:112` | allowed, see below |

The last one is the only judgement call. The natural rewrite does not compile:

```
error[E0382]: use of moved value: `file`
   --> src/producer.rs:96:46
```

`file` is an `Option<&mut impl Read>` and so not `Copy`, and a value moved in a match guard counts as moved for the whole match, so the three content checks in `handle_file` have to stay in the arm bodies. I put an `#[allow(clippy::collapsible_match)]` on the function with that reason written next to it rather than restructure the match. Happy to do it differently if you would rather.

One thing I deliberately left out: an older clippy (1.95 here) additionally reports `collapsible_match` at `src/parser.rs:742` - the current CI clippy does not. It cannot become a guard either, because the inner condition uses `?`. I left `parser.rs` untouched so this does not conflict with #1515 and #1516, which both change that file. Say the word and I will fold it in.

Checked locally: `cargo clippy --bins --tests --examples --all -- -D rust_2018_idioms -D warnings` is clean apart from that one `parser.rs` line, `cargo fmt -- --check` passes, and `cargo test --lib` is unchanged at 133 passed (the four `llvm_tools` failures here need `llvm-profdata` and fail identically on `master`).

### DIFF
--- a/src/cobertura.rs
+++ b/src/cobertura.rs
@@ -65,8 +65,8 @@ impl CoverageStats {
         let lines_valid = lines.len() as f64;
 
         let branches: Vec<Vec<Condition>> = lines
-            .into_iter()
-            .filter_map(|(_, l)| match l {
+            .into_values()
+            .filter_map(|l| match l {
                 Line::Branch { conditions, .. } => Some(conditions),
                 Line::Plain { .. } => None,
             })
--- a/src/path_rewriting.rs
+++ b/src/path_rewriting.rs
@@ -417,18 +417,10 @@ pub fn rewrite_paths(
             }
 
             match filter_option {
-                Some(true) => {
-                    if !is_covered(&result) {
-                        return None;
-                    }
-                }
-                Some(false) => {
-                    if is_covered(&result) {
-                        return None;
-                    }
-                }
-                None => (),
-            };
+                Some(true) if !is_covered(&result) => return None,
+                Some(false) if is_covered(&result) => return None,
+                _ => (),
+            }
 
             Some((abs_path, rel_path, result))
         });
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -50,6 +50,11 @@ impl Archive {
             .push(self);
     }
 
+    // The three content checks below read from `file`, which is an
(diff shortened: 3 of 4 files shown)


Eljees:
Green now - 10/10, `Lint` included.

Two follow-up commits were needed, both from lints that only appear once the previous error stops the build early:

- `redundant reference in panic! argument` in `tests/test.rs:584`, which the library failing to compile had been hiding.
- Then, from my first attempt at that one: I inlined the binding as `panic!("Failed to copy {gcda_zip_path:?}")`, which is wrong here - the crate is `edition = "2018"`, where implicit format-args capture does not exist, so the string is taken literally and rustc reports an unused placeholder. Dropping just the redundant `&` and keeping `panic!("Failed to copy {:?}", gcda_zip_path)` is what the lint actually wanted.

Worth flagging the reason for the churn: the clippy CI installs is newer than any toolchain I can reach, so I cannot reproduce its exact lint set locally and had to find those two through CI. If you have a preference for pinning the toolchain - a `rust-toolchain.toml`, or `--locked` clippy in the hook - that would stop this recurring on the next bump, and I am happy to send it separately.

Once this lands, #1515 and #1516 only need `master` merged in to go green; the same goes for the dependabot PRs.

marco-c (approved):

PR #1521, opened 2026-09-09
Add GCNO header b"11B" to the LLVM auto-detection

nicolas-guichard:
This is the default (and minimum) version since Clang 20 [1].

Based on our test corpus the default for GCC 9 is b"39A" and the default for GCC 10 is b"20B".

[1] https://github.com/llvm/llvm-project/commit/82fecab85ae2d72ffac0e44749d99f12d6f71cc0


marco-c:
Can we add a test file that exercises this?

nicolas-guichard:
We already have one, which is why the tests were failing at first.

marco-c (approved):

PR #1530, opened 2026-09-18
Bug 2033336 - Normalize paths in rename_external_files

nicolas-guichard:
The rename_external_files function used to output paths such as:
```
/builds/worker/workspace/obj-build/modules/libpref/./../../../../checkouts/gecko/modules/libpref/Preferences.cpp
```

This was eventually normalized to:
```
/builds/worker/checkouts/gecko/modules/libpref/Preferences.cpp
```

But that normalization happened too late, after the removal of the
prefix set with the -p option.

We need to normalize these somewhere between rename_external_files and
the remove_prefix call in rewrite_paths. I chose to do it here to limit
potential side effects to a minimum.


marco-c:
path rewriting is a mess, I often wonder if we shouldn't remove all that and rewrite it from scratch in a much simpler way with way fewer options

marco-c (approved):

